PR #429, opened 2014-12-25
Refactor tasks and services to accept only options with url

### LINKED ISSUE #420: Refactor tasks and services to accept only options with URL
I had a conversation with @jgravois about this today so I will attempt to transcribe the notes here.

Initializing Tasks and Services with the `(url, options)` pattern is starting to cause problems that look like horrible kludges in downstream plugins like the geocoder.

This proposed code by @jgravois is pretty indicative of the problem. https://github.com/jgravois/esri-leaflet-geocoder/blob/empty-constructor/src/Tasks/Geocode.js#L28-L41. That code has to check for:

1. passing a Service as the first param,
2. passing `url` and `options`
3. passing `options` and using the default URL

In short we shouldn't have to write code like this. Its much cleaner to make `url` an option that can be passed in. This means that the first parameter is always either

1. an options object with `url`, or no `url` to use a default URL
2. a service, in which case we only need to check for the presence of `request` and switch accordingly

This is a major breaking change but we are still in a release candidate stage of Esri Leaflet now is the time to do this to simplify downstream plugins that want to use a default for URL.

jgravois:
attempt to resolve #420

this is my first pass at refactoring Service and Task to expect `{url: 'myUrl'}`

i've manually tested about a dozen samples that demo all the different classes in the API and they all behave as expected
i also refactored the tests (which helped me catch a few bugs along the way).
all new tests are passing.

i'll wait to squash until everyone's happy with the code.

### DIFF
--- a/spec/Layers/BaseMapLayerSpec.js
+++ b/spec/Layers/BaseMapLayerSpec.js
@@ -49,7 +49,7 @@ describe('L.esri.Layers.BasemapLayer', function () {
     for (var i = 0, len = testmaps.length; i < len; i++) {
       var name = testmaps[i];
       expect(L.esri.basemapLayer(name)).to.be.instanceof(L.esri.Layers.BasemapLayer);
-      expect(L.esri.basemapLayer(name)._url).to.eql(L.esri.BasemapLayer.TILES[name].urlTemplate);
+      expect(L.esri.basemapLayer(name)._url).to.eql(L.esri.BasemapLayer.TILES[name].options.url);
     }
   });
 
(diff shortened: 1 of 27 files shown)


patrickarlt:
@alaframboise @jgravois I keep coming back to how Leafelt and the JS API both work. The first parameter in a layer is always something essential. GeoJSON, coordinates, a url, a template. Moving the Tasks and Services to this makes sense because it simplifies downstream plugins like routing and geocoding but moving layers to this pattern is just a big breaking change.

I've merged the other PR on this so I think I was to go forward with the approach with URL as its own parameter.

jmfolds:
With v1.0.0-rc.5, dynamic layers still take a url as the first parameter. After instantiating a layer, the url property is now stored on layer.options rather than on the layer itself, is this the expected behavior?

Thanks for your help!

jgravois:
@jmfolds i did this so we could pass all relevant options to L.Util.[setOptions()](http://leafletjs.com/reference.html#util-setoptions) in one fell swoop (see [here](https://github.com/Esri/esri-leaflet/blob/master/src/Layers/DynamicMapLayer.js#L14-L17)), but i can't say thats crucial.

you think its sufficient to document the breaking change a little more clearly?  it just dawned on me that we are publicizing layer properties in the API reference whatsoever.  Leaflet does this too, and i've never been sure why, with the exception of there being less to maintain.

jmfolds:
@jgravois I was mainly checking to make sure this was what was intended. I think a little clarity on the breaking change wouldn't hurt and would be sufficient. Thanks.

jgravois:
@jmfolds added 74cb11c349652500320744187b13e38f80df85db
thanks for pointing out the issue.

jmfolds:
@jgravois No problem. Thanks for your help, and for this project!

PR #432, opened 2014-12-30
modify all Task and Service class constructors to take 'url' within an options object

jgravois:
incorporated pat's suggestion of leaving constructor for layers alone.

initially i tried to isolate the change to L.esri.Tasks.Task and L.esri.Services.Service only, but it seemed necessary to mimic in other tasks and services as well because they don't necessarily all provide a hook to override/inherit the initialize method and pass constructor options differently (for example: [L.esri.Tasks.Identify](https://github.com/Esri/esri-leaflet/blob/master/src/Tasks/Identify.js).  its totally possible i'm missing something easy, so i won't be offended if thats pointed out to me here.

either way, take a close look at new logic [here](https://github.com/jgravois/esri-leaflet/compare/Esri:master...jgravois:task-service-constructor?expand=1#diff-a7be8ad7c53337df7dc83d076f320957R13).  what i did seems cheesy, but it was the best approach i could come up with.

* modified all Service and Task tests to account for design change
* commented out min/maxZoom test which is currently failing 
* updated variable name for urls in tests to be more descriptive
* fixed typo in test filename
* all new tests passing (using `--force`)


patrickarlt:
@jgravois forget about the `TileLayer` I misread it. Logic for setting options looks good.

PR #437, opened 2015-01-12
Update location of feature layer

rowanwins:
Changed feature layer from 
http://services.arcgis.com/uCXeTVveQzP4IIcx/arcgis/rest/services/bike_rte/FeatureServer/0
to
http://services.arcgis.com/uCXeTVveQzP4IIcx/ArcGIS/rest/services/Bike_Routes/FeatureServer/0

PR #444, opened 2015-01-22
miscellaneous documentation fixes

jgravois:
this is a collection of about half a dozen miscellaneous documentation fixes, many of which were merged previously (via PRs against `gh-pages`)

highlights:
url fix in sample #437
addition of Task documentation page #433
renaming of Utils page to match `L.esri.Util` #357
some copy edits for clarity #395
fix url in background map #387

i've built the source and confirmed that the static pages look correct, but i'm at a bit of a loss to understand how broken references to "/js/esri-leaflet.js" in the examples src (see [here](https://github.com/Esri/esri-leaflet/blob/master/site/source/layouts/example.hbs#L28)) are replaced with links to the CDN (see [here](https://github.com/Esri/esri-leaflet/blob/gh-pages/examples/arcgis-online-auth.html#L38)) in the live site.  if i can wrap my head around that piece i would be able to perform additional testing to make sure everything looks perfect prior to merging.


jgravois:
ive got a system now to build locally and have confirmed that the updated site is looking good.

jgravois:
i have laid down several more commits on this PR over the last couple days:

fixed bugs w/ current geocoding samples
bumped copyright
added new samples to introduce the renderers and gp plugins
moved PDX org services to https
updated CDN to leverage jsdelivr (since it supports both http and https)

next time i promise not to clutter up one PR with so many unrelated changes/fixes.

PR #457, opened 2015-02-13
better check to make sure logo hasn't already been removed

### LINKED ISSUE #455: When switching basemaps, unexpected errors occur
I apologize if this has already been brought up. I ran into an issue creating a basemap switcher. The issue occurs after the basemap is switched multiple times. 
For example see: http://esri.github.io/esri-leaflet/examples/switching-basemaps.html 
The issue appears when switching from these basemaps:
-Oceans
-Gray
-Dark Gray
-Imagery
-Shaded Relief

I am using Google Chrome to test this, and it appears to be occuring [here](https://github.com/Esri/esri-leaflet/blob/master/src/Layers/BasemapLayer.js#L231)

` Uncaught TypeError: Failed to execute 'removeChild' on 'Node': parameter 1 is not of type 'Node'.
leaflet.js:9 
o.Control.o.Class.extend.removeFrom
leaflet.js:9 
o.Map.include.removeControl                        
BasemapLayer.js:231 
a.Layers.BasemapLayer.L.TileLayer.extend.onRemove
leaflet.js:6 
o.Map.o.Class.extend.removeLayerswitching-basemaps.html:205 
setBasemap
switching-basemaps.html:218 (anonymous function)`

jgravois:
resolves #455.  previously we checked for `this._logo`, prior to attempting to remove it from the map, but this is still an object (without a corresponding DOM node)

the new approach of checking for the presence of `this._logo._container` resolves problems that manifested when a basemap and corresponding label layer were removed from the map simultaneously.  

i've added a new test to trap problems like this in the future as well.

### DIFF
--- a/src/Layers/BasemapLayer.js
+++ b/src/Layers/BasemapLayer.js
@@ -228,7 +228,7 @@
     },
     onRemove: function(map){
       // check to make sure the logo hasn't already been removed
-      if(this._logo._container){
+      if(this._logo && this._logo._container){
         map.removeControl(this._logo);
         map._hasEsriLogo = false;
       }
--- a/src/Layers/BasemapLayer.js
+++ b/src/Layers/BasemapLayer.js
@@ -227,7 +227,8 @@
       map.on('moveend', this._updateMapAttribution, this);
     },
     onRemove: function(map){
-      if(this._logo){
+      // check to make sure the logo hasn't already been removed
+      if(this._logo._container){
         map.removeControl(this._logo);
         map._hasEsriLogo = false;
       }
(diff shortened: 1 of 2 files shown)


jgravois:
not sure what's up with Travis.  all tests are passing locally.

src/Layers/BasemapLayer.js
```diff
@@ -227,7 +227,8 @@
       map.on('moveend', this._updateMapAttribution, this);
     },
     onRemove: function(map){
-      if(this._logo){
+      // check to make sure the logo hasn't already been removed
+      if(this._logo._container){
```
patrickarlt: This should probally be `if(this._logo && this._logo._container) {`

src/Layers/BasemapLayer.js
```diff
@@ -227,7 +227,8 @@
       map.on('moveend', this._updateMapAttribution, this);
     },
     onRemove: function(map){
```
jgravois: sounds good.  I'll fix and merge.

Issue #463, opened 2015-02-24
Leaflet is silently replaced by Esri-leaflet in an envivonment with browserify and a special leaflet version.


burmisov:
Say I use my own Leaflet library version based on master branch or some other patched one.
For instance, we add it to browserify bundle like so:
```js
require('my-special-leaflet'); // Creates window.L with my own Leaflet version (as a side-effect).
```
Then I try to add Esri-leaflet to play:
```js
require('esri-leaflet'); // Supposed to patch window.L and put window.L.esri there.
```
But as far as I get it, given this environment, Esri-leaflet bootstrap code (which checks for amd/commonjs/...) replaces window.L with its own bundled Leaflet version (0.7.2 as of today), which is not a behavior I would expect. Moreover, I can't think of a proper workaround to force use my own Leaflet version.

jgravois:
i can't reproduce this error.  check out the console in [this](https://dl.dropboxusercontent.com/u/59331579/js/browserify-test/index.html) sample app.

in my package.json, i am loading 0.7.0 explicitly
```
{
  "name": "browserifyTest",
  "main": "index.js",
  "author": "John Gravois <jgravois@esri.com> (http://johngravois.com)",
  "license": "Apache 2.0",
  "readmeFilename": "README.md",
  "dependencies": {
    "leaflet": "0.7.0",
    "esri-leaflet": "1.0.0-rc.5"
  }
}
```
this is my node app
```
require('leaflet');
require('esri-leaflet');

var mapDiv = document.createElement('div')
mapDiv.setAttribute('id', 'mapDiv');
mapDiv.setAttribute('style', 'position: fixed; top: 0; bottom: 0; left: 0; right: 0;')
document.body.appendChild(mapDiv);

var map = L.map('mapDiv').setView([42.36, -71.06], 13);
L.esri.basemapLayer('Gray').addTo(map);

console.log(L.version);
console.log(L.esri.VERSION);
```
i'm happy to chat more if you can provide more information.  at the very least, its helpful to know that its time for us to bump esri-leaflets internal leaflet dependency to `0.7.3`

jgravois:
the key here is that you have to `require('leaflet')` whether you are working with an older tagged version or any other specific patch.  this is what let's esri-leaflet know to rely on the module you've already loaded for its own dependency.

burmisov:
Thanks for a prompt reply, John.
[...]

PR #467, opened 2015-03-01
Make leaflet a 'devDependency'

jgravois:
resolves problems in projects that use browserify and attempt to load a patched version of Leaflet (see #463).
follows the [pattern](https://github.com/Leaflet/Leaflet.toolbar/blob/master/package.json#L9-L36) of other plugins in the Leaflet organization.

also added a 'repository' tag to appease an `npm install` nag.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,7 @@
 {
   "name": "esri-leaflet",
   "version": "1.0.0-rc.5",
+  "repository": "Esri/esri-leaflet",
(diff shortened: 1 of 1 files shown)


patrickarlt:
While this doesn't break browserify support (I checked) it feels really weird to me because Leaflet is a dependency in production not just development when using browserify and I feel we need to declare it as such since I'm not sure if this is something that might change in the future.

PR #484, opened 2015-03-17
trim height of export requests when necessary

### LINKED ISSUE #450: Dynamic Layer rendering issue on Zoom level 1 and 2
Please see your example as reference:  http://esri.github.io/esri-leaflet/examples/simple-dynamic-map-layer.html.  Zoom all the way out and keep your eye on Alaska.  When you zoom all the way out, the layer becomes misplaced/pushed down.  We have noticed this on all dynamic layers that we are using as well.  I apologize if this is an existing issue, but I don't see this anywhere else.  Just need to know if this issue has been or will be addressed.  Thanks.

jgravois:
resolves #450

i was able to correct the distortion seen in DynamicMapLayer overlays in scenarios where whitespace surrounds the map by introducing a quick check into `_buildExportParams` to compare the height of the div to the top and bottom y values of the actual map (in screen coordinates) and trimming if appropriate.

confirmed that display behavior is improved for web mercator basemaps, even when L.tileLayer noWrap is set to true
confirmed that display behavior is appropriate in leaflet-1.0 branch as well (with the exception of the fact that the area below 85deg south isn't cropped appropriately)

happy to include relevant tests prior to merge, just wanted to make sure everyone's happy with the methodology first.
cc @nixta

### DIFF
--- a/src/Layers/DynamicMapLayer.js
+++ b/src/Layers/DynamicMapLayer.js
@@ -91,12 +91,11 @@ EsriLeaflet.Layers.DynamicMapLayer = EsriLeaflet.Layers.RasterLayer.extend({
     var sw = this._map.options.crs.project(bounds._southWest);
 
     //ensure that we don't ask ArcGIS Server for a taller image than we have actual map displaying
-    var total = this._map.getSize();
     var top = this._map.latLngToLayerPoint(bounds._northEast);
     var bottom = this._map.latLngToLayerPoint(bounds._southWest);
 
-    if (top.y > 0 || bottom.y < total.y){
-      size.y = total.y - top.y - (total.y - bottom.y);
+    if (top.y > 0 || bottom.y < size.y){
+      size.y = bottom.y - top.y;
     }
 
     var params = {
--- a/src/Layers/DynamicMapLayer.js
+++ b/src/Layers/DynamicMapLayer.js
@@ -90,6 +90,15 @@ EsriLeaflet.Layers.DynamicMapLayer = EsriLeaflet.Layers.RasterLayer.extend({
     var ne = this._map.options.crs.project(bounds._northEast);
     var sw = this._map.options.crs.project(bounds._southWest);
 
+    //ensure that we don't ask ArcGIS Server for a taller image than we have actual map displaying
+    var total = this._map.getSize();
+    var top = this._map.latLngToLayerPoint(bounds._northEast);
+    var bottom = this._map.latLngToLayerPoint(bounds._southWest);
+
+    if (top.y > 0 || bottom.y < total.y){
+      size.y = total.y - top.y - (total.y - bottom.y);
+    }
+
     var params = {
       bbox: [sw.x, sw.y, ne.x, ne.y].join(','),
       size: size.x + ',' + size.y,


src/Layers/DynamicMapLayer.js
```diff
@@ -90,6 +90,15 @@ EsriLeaflet.Layers.DynamicMapLayer = EsriLeaflet.Layers.RasterLayer.extend({
     var ne = this._map.options.crs.project(bounds._northEast);
     var sw = this._map.options.crs.project(bounds._southWest);
 
+    //ensure that we don't ask ArcGIS Server for a taller image than we have actual map displaying
+    var total = this._map.getSize();
+    var top = this._map.latLngToLayerPoint(bounds._northEast);
+    var bottom = this._map.latLngToLayerPoint(bounds._southWest);
+
+    if (top.y > 0 || bottom.y < total.y){
+      size.y = total.y - top.y - (total.y - bottom.y);
```
nixta: Maybe reduce this to `size.y = bottom.y - top.y`? Then you can also get rid of `total` completely.

src/Layers/DynamicMapLayer.js
```diff
@@ -90,6 +90,15 @@ EsriLeaflet.Layers.DynamicMapLayer = EsriLeaflet.Layers.RasterLayer.extend({
     var ne = this._map.options.crs.project(bounds._northEast);
     var sw = this._map.options.crs.project(bounds._southWest);
 
+    //ensure that we don't ask ArcGIS Server for a taller image than we have actual map displaying
+    var total = this._map.getSize();
```
nixta: This is a duplicate of `size` declared earlier.

src/Layers/DynamicMapLayer.js
jgravois: i'm still using `total` in the check above, but :+1: for simpler maths!

src/Layers/DynamicMapLayer.js
jgravois: i did that on purpose thinking it made the comparison easier to read, but i'm fine with it the other way too.

nixta: As it is I think it reads better without total, but I'm no Leaflet developer.

src/Layers/DynamicMapLayer.js
```diff
@@ -90,6 +90,14 @@ EsriLeaflet.Layers.DynamicMapLayer = EsriLeaflet.Layers.RasterLayer.extend({
     var ne = this._map.options.crs.project(bounds._northEast);
     var sw = this._map.options.crs.project(bounds._southWest);
 
+    //ensure that we don't ask ArcGIS Server for a taller image than we have actual map displaying
+    var top = this._map.latLngToLayerPoint(bounds._northEast);
+    var bottom = this._map.latLngToLayerPoint(bounds._southWest);
+
+    if (top.y > 0 || bottom.y < size.y){
```
jgravois: looking at this again, i guess we could even skip the check entirely.  
if the map fills up the whole div, `bottom.y - top.y` still returns the height we want.

patrickarlt: Lets leave the check in there just for safety.

patrickarlt:
Looks good!

PR #491, opened 2015-03-23
added an 'alt' tag to L.esri.Logo

### LINKED ISSUE #490: Alt text for "Powered by Esri" logo
The "Powered by esri" logo in the bottom right has no alt text, which causes accessibility checker errors. It's easy enough to hack the text in with jQuery after the map loads with something like this: 

`$('.esri-leaflet-logo img').attr('alt','Powered by Esri logo');`

but it would be nice if it just happened out of the box. That would provide an added benefit of consistent alt text across sites.

Probably not the biggest issue in the world, but hopefully a simple fix.

jgravois:
resolves #490


patrickarlt:
This should probably just be "Powered by Esri" which is the full text. It is not really a logo.

jgravois:
agreed.

PR #492, opened 2015-03-23
fixes for problems in the editing sample

jgravois:
#### included fixes:
- [x] the dialog which assists with editing attributes doesn't disappear after a feature has been deleted.  (it would be preferable if it were not displayed at all in that scenario)
- [x] its not possible to modify the geometry of a feature after deleting something else.
- [x] 'Pedetrian' is misspelled
- [ ] clicking a feature and then clicking the map triggers an `updateFeature` call, even though the feature hasn't changed
- [x] each time an edit is passed, the featureLayer `load` event is triggered and another set of queries to the service fire.  perhaps this is by design, but it seems like it would be preferable if we could update the L.esri.FeatureLayer and L.FeatureGroup for the Draw plugin more surgically.

not sure if its worth doing anything about the other two...

### DIFF
--- a/src/Layers/FeatureLayer/FeatureManager.js
+++ b/src/Layers/FeatureLayer/FeatureManager.js
@@ -396,6 +396,7 @@
     },
 
     addFeature: function(feature, callback, context){
+      //still need to pass 'undefined' as a placeholder, not sure how to fix
       this._getMetadata(L.Util.bind(function(undefined, error, metadata){
         this._service.addFeature(feature, function(error, response){
           if(!error){
--- a/src/Layers/FeatureLayer/FeatureManager.js
+++ b/src/Layers/FeatureLayer/FeatureManager.js
@@ -385,32 +385,33 @@
 
     _getMetadata: function(callback, context){
       if(this._metadata){
-        callback(undefined, this._metadata, this);
+        var error;
+        callback(context, error, this._metadata);
       } else {
-        this.metadata(L.Util.bind(function(error, response, context) {
+        this.metadata(L.Util.bind(function(error, response) {
           this._metadata = response;
-          callback(error, this._metadata, this);
+          callback(context, error, this._metadata);
         }, this));
       }
     },
 
     addFeature: function(feature, callback, context){
-      this._getMetadata(function(error, metadata, context){
-        context._service.addFeature(feature, function(error, response){
+      this._getMetadata(L.Util.bind(function(undefined, error, metadata){
+        this._service.addFeature(feature, function(error, response){
           if(!error){
             // assign ID from result to appropriate objectid field from service metadata
             feature.properties[metadata.objectIdField] = response.objectId;
 
             // we also need to update the geojson id for createLayers() to function
             feature.id = response.objectId;
-            context.createLayers([feature]);
+            this.createLayers([feature]);
           }
           if(callback){
             callback.call(context, error, response);
           }
-        }, context);
+        }, this);
         return this;
-      }, this);
+      }, this), this);
     },
 
     updateFeature: function(feature, callback, context){
--- a/src/Layers/FeatureLayer/FeatureManager.js
+++ b/src/Layers/FeatureLayer/FeatureManager.js
@@ -386,7 +386,19 @@
     addFeature: function(feature, callback, context){
       this._service.addFeature(feature, function(error, response){
         if(!error){
-          this.refresh();
+          /*
+          this code requires the developer supply an 'idAttribute' in the constructor options
+          currently, 'this._service.options.idAttribute' is hardcoded to 'OBJECTID'
+          */
+          if (this.options.idAttribute){
+            feature.properties[this.options.idAttribute] = response.objectId;
+          }
+          else {
+            feature.properties.OBJECTID = response.objectId;
+          }
+          // we also need to update the geojson id for createLayers() to function
+          feature.id = response.objectId;
+          this.createLayers([feature]);
         }
         if(callback){
           callback.call(context, error, response);
@@ -398,7 +410,8 @@
     updateFeature: function(feature, callback, context){
       return this._service.updateFeature(feature, function(error, response){
         if(!error){
-          this.refresh();
+          this.removeLayers([feature.id], true);
+          this.createLayers([feature]);
         }
         if(callback){
           callback.call(context, error, response);
@@ -416,6 +429,7 @@
         }
       }, this);
     }
+
   });
 
   /**
--- a/src/Layers/FeatureLayer/FeatureManager.js
+++ b/src/Layers/FeatureLayer/FeatureManager.js
@@ -383,28 +383,34 @@
       return this._service.query();
     },
 
+    _getMetadata: function(callback, context){
+      if(this._metadata){
+        callback(undefined, this._metadata, this);
+      } else {
+        this.metadata(L.Util.bind(function(error, response, context) {
+          this._metadata = response;
+          callback(error, this._metadata, this);
+        }, this));
+      }
+    },
+
     addFeature: function(feature, callback, context){
-      this._service.addFeature(feature, function(error, response){
-        if(!error){
-          /*
-          this code requires the developer supply an 'idAttribute' in the constructor options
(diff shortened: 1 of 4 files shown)


patrickarlt:
> each time an edit is passed, the featureLayer load event is triggered and another set of queries to the service fire. perhaps this is by design, but it seems like it would be preferable if we could update the L.esri.FeatureLayer and L.FeatureGroup for the Draw plugin more surgically.

Agreed that we could do this WAY better Its probably a matter of a few tweaks in https://github.com/Esri/esri-leaflet/blob/master/src/Layers/FeatureLayer/FeatureManager.js#L386-L419.

* `addFeature` calls `this. createLayers([feature]);`
* `updateFeature` calls `this.removeLayers([featureId], true);` to remove the feature from the map and the cache then calls `this. createLayers([feature]);` to create the updated feature

Full refresh is lazy and I was lazy for doing it.

jgravois:
i've laid down another commit which gets us a lot of the way toward updating the clientside graphics after an edit without requerying the service.

some things to consider:
1. in addition to the changes you suggested, it was necessary to interrogate `FeatureLayer` to find out the name of the objectid field in use and populate that attribute value using the server response.  since we don't currently have a dependable reference to it and i know you don't want to fire off a metadata() request when initializing the class, i took the liberty of creating a new constructor option so that developers to declare it manually.  i'm happy to pursue another approach if you like.
2. its also necessary to pass the same attribute value as the geoJson 'id' in order to handle redrawing the clientside graphic.
3. the service we are using in our editing sample right now has a lot of different fields that one might think could be used to index (*FID, OBJECTID, GEODB_OID*), but `FID` is actually the one that needs to be populated as a feature property in order to be updatable later (even though its type is `esriFieldTypeInteger` [here](http://services.arcgis.com/rOo16HdIMeOBI4Mb/arcgis/rest/services/PDX_Pedestrian_Districts/FeatureServer/0?f=json).  super confusing!
4. my changes cause [these](https://github.com/Esri/esri-leaflet/blob/master/spec/Layers/FeatureLayer/FeatureManagerSpec.js#L635-L696) two tests to fail (declaring that spy.callCount is now 0), but it seems like a false positive, because i can clearly see in my own test app that the appropriate requests are fired.
5. i wrote up a new `deleteFeatures()` method and updated the test app after noticing that we were passing multiple simultaneous requests to the service rather than an array of objectIds when people *save* deletes in the draw tool which i will submit in another PR.  

in general these changes dramatically cut down on network chatter when editing.

patrickarlt:
> since we don't currently have a dependable reference to it and i know you don't want to fire off a metadata() request when initializing the class, i took the liberty of creating a new constructor option so that developers to declare it manually. i'm happy to pursue another approach if you like.

I'm assuming this was for `updateFeature`. Another way to accomplish this would be to request the metadata inside `updateFeature`.

```js
_getMetadata: function(callback){
  if(this._metadata){
    callback(undefined, this._metadata);
  } else {
    this.metadata(L.Util.bind(function(error, response) {
      this._metadata = response;
      callback(error, this._metadata)
    }, this));
  }
},

updateFeature: function(feature, callback, context) {
  this._getMetadata(function(error, metadata){
    feature = EsriLeaflet.Util.geojsonToArcGIS(feature, this.options.idAttribute);
    this.post('updateFeatures', {
      features: [feature]
    }, function(error, response){
      var result = (response && response.updateResults) ? response.updateResults[0] : undefined;
      if(callback){
        callback.call(context, error || response.updateResults[0].error, result);
      }
    }, context);
  });
}
```

because `updateFeature` is async anyway it just adds a small delay to the first operation. I just don't want that first operation to be layer loading.

jgravois:
> request the metadata inside...
i can't believe i didn't think of that myself.

i've laid down another commit and inserted the logic you suggested within `Layers.FeatureManager.addFeature()` as opposed to `Services.FeatureLayer.updateFeature()` because that is our opportunity to snag `objectids`.  (ie: because the server returns the unique value in the response when new features are inserted successfully) 

#### more considerations
1. the code works, but i get a little mixed up between passing around `context` and `this`
2. the two tests i mentioned previously are still failing

patrickarlt:
`context` vs `this` sucks but its a leaflet pattern we are kinda stuck with.

Here is how I keep them straight.

* `context` is always what you should `call()` the users callback with. Never use to for anything else.
* When nesting lots of callbacks where you need to pass `this` around, pass `this` as the last parameter or use `L.Util.bind(this, function(){ ... })`

This is more then a little gross and it almost makes me want to drop context everywhere in the designs. I'll review the changes though.

src/Layers/FeatureLayer/FeatureManager.js
```diff
@@ -383,22 +383,41 @@
       return this._service.query();
     },
 
+    _getMetadata: function(callback, context){
+      if(this._metadata){
+        callback(undefined, this._metadata, this);
+      } else {
+        this.metadata(L.Util.bind(function(error, response, context) {
+          this._metadata = response;
+          callback(error, this._metadata, this);
```
patrickarlt: This should be `callback.call(context, error, this._metadata)` don't pass `context` to the callback function. Using `call` with context as the first param mean that `this` inside the callback function will equal `context`.

src/Layers/FeatureLayer/FeatureManager.js
```diff
@@ -383,22 +383,41 @@
       return this._service.query();
     },
 
+    _getMetadata: function(callback, context){
+      if(this._metadata){
+        callback(undefined, this._metadata, this);
```
patrickarlt: This should be `callback.call(context, error, this._metadata)` don't pass `context` to the callback function. Using `call` with context as the first param mean that `this` inside the callback function will equal `context`.

src/Layers/FeatureLayer/FeatureManager.js
```diff
@@ -383,22 +383,41 @@
...
+            feature.properties[metadata.objectIdField] = response.objectId;
+
+            // we also need to update the geojson id for createLayers() to function
+            feature.id = response.objectId;
+            context.createLayers([feature]);
+          }
+          if(callback){
+            callback.call(context, error, response);
+          }
+        }, context);
```
patrickarlt: With the above changes to the `_getMetadata()` method you should pass `this` here not context.

src/Layers/FeatureLayer/FeatureManager.js
```diff
@@ -383,22 +383,41 @@
...
+
     addFeature: function(feature, callback, context){
-      this._service.addFeature(feature, function(error, response){
-        if(!error){
-          this.refresh();
-        }
-        if(callback){
-          callback.call(context, error, response);
-        }
+      this._getMetadata(function(error, metadata, context){
```
patrickarlt: Now that `context` doesn't get passed through you should remove it from here.

src/Layers/FeatureLayer/FeatureManager.js
```diff
@@ -383,22 +383,41 @@
...
+        context._service.addFeature(feature, function(error, response){
```
patrickarlt: You can now use `this._service.addFeature`.

src/Layers/FeatureLayer/FeatureManager.js
```diff
@@ -383,22 +383,41 @@
...
+          if(!error){
+            // assign ID from result to appropriate objectid field from service metadata
+            feature.properties[metadata.objectIdField] = response.objectId;
+
+            // we also need to update the geojson id for createLayers() to function
+            feature.id = response.objectId;
+            context.createLayers([feature]);
```
patrickarlt: `this.createLayers`

jgravois:
thanks for taking the time to explain.  it *almost* makes sense to me now. :smile:

i incorporated your suggestions and get appropriate behavior and for the most part the code is more readable and consistent, but i'm still left with one pretty insane `this` rat's nest at the bottom of `FeatureManager.addFeature()` [here](https://github.com/jgravois/esri-leaflet/blob/editing-sample/src/Layers/FeatureLayer/FeatureManager.js#L412-L414) that i'm sure could be improved substantially...

patrickarlt:
@jgravois is this PR ready to merge now?

PR #493, opened 2015-03-25
bump renderer sample to beta.3

jgravois:
swapped in a new smart mapping ArcGIS Online feature service too


kneemer:
@patrickarlt Is the feature layer in the sample one of your test layers or will it be around for some time?

patrickarlt:
@kneemer I'll make sure it is.

PR #494, opened 2015-03-25
fix cleanUrl regex so that spaces in folder names arent trimmed

jgravois:
i noticed this morning that L.Util.cleanUrl incorrectly trims spaces inside url folder names.

```javascript
L.esri.Util.cleanUrl('http://utility.arcgisonline.com/arcgis/rest/services/Utilities/PrintingTools/GPServer/Export Web Map Task/execute  ');
//Export Web Map Task becomes ExportWebMapTask (and induces 400 errors in requests)
>>"http://utility.arcgisonline.com/arcgis/rest/services/Utilities/PrintingTools/GPServer/ExportWebMapTask/execute/"
```
i've fixed using another regex i found on the internet.

### DIFF
--- a/spec/UtilSpec.js
+++ b/spec/UtilSpec.js
@@ -44,6 +44,11 @@ describe('L.esri.Util', function () {
     expect(url).to.equal('http://arcgis.com/');
   });
 
+  it('shouldnt trim spaces in the middle', function(){
+    var url = L.esri.Util.cleanUrl('   http://arcgis.com/cool folder/anotherfolder ');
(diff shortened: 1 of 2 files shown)


patrickarlt:
@jgravois I think you could replace using 2 calls to replace with this.

```js
url.replace(/^\s+|\s+$|\A\s+|\s+\z/g, '');
```

It's a little silly but it should only match whitespace at the start/end of the string/line. Here is a link to test it http://rubular.com/r/PgHbPEL6EO

jgravois:
sweet.  fixed, ran grunt tests and confirmed manually everything was good, rebased and merging now.

PR #502, opened 2015-04-08
Update FAQ links in the README

whymarrh:
Small fix: I updated the links in the README with the correct links.


jgravois:
thanks for this catch @whymarrh.

would you be willing to do a find and replace to fix up all the FAQ links in the readme to update all of them to use relative links?  maybe im being OCD, but it seems like this would be best.

was:
link](https://github.com/Esri/esri-leaflet/wiki/FAQ#when-will-feature-x-get-done)
should be:
link](wiki/FAQ#when-will-feature-x-get-done)

whymarrh:
Relative links might introduce a few unintended side-effects:

1. Links wouldn't work outside of GitHub (e.g. local Markdown viewer)
2. Links wouldn't work when viewing the [README file directly](https://github.com/Esri/esri-leaflet/blob/master/README.md)

    (That is, relative to `https://github.com/Esri/esri-leaflet/blob/master/README.md`, `wiki/FAQ` would be `https://github.com/Esri/esri-leaflet/blob/master/wiki/FAQ`)

jgravois:
i don't really consider the first behavior a consequence.
with regard to the second, i think using `/wiki/..` instead of `wiki/` would do the trick.
i don't have my heart set on it one way or the other, i just kind of felt like *modularity > hardcoding*

whymarrh:
How does that look?

jgravois:
thanks for that.  if you aren't already tired of humoring me, please give it a quick [squash](http://blog.steveklabnik.com/posts/2012-11-08-how-to-squash-commits-in-a-github-pull-request) when you have a moment and I'll merge.

whymarrh:
No worries, though it should be just the one commit there already

jgravois:
:+1:

PR #510, opened 2015-04-15
added deleteFeatures methods

jgravois:
i've added a couple methods so that clients can pass an array of objectIds in a single request to the server.

```javascript
map.on('draw:deleted', function(e) {
    var delArray = [];
    e.layers.eachLayer(function(layer) {
      var id = layer.feature.id;
      delArray.push(id);
    });
    pedestrianDistricts.deleteFeatures(delArray, function(error, response) {
      console.log(error, response);
    });
});
```
to do:
- [x] - write tests
- [ ] - update our sample app to use the new method
- [x] - update the doc

### DIFF
--- a/site/source/pages/api-reference/layers/feature-layer.md
+++ b/site/source/pages/api-reference/layers/feature-layer.md
@@ -277,7 +277,7 @@ featureLayer.query()
                 Adds a new feature to the feature layer. this also adds the feature to the map if creation is successful.
                 <ul>
                     <li>Requires authentication as a user who has permission to edit the service in ArcGIS Online or the user who created the service.</li>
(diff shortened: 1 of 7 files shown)


patrickarlt:
@jgravois looks like the tests are doing their usual failures for unrelated reasons. Could you finish this up today? I could fix the tests (again) and do a release tomorrow.

jgravois:
ok.  ready to merge now.  im going to wait until we tag to update the editing sample to use the new method.  (i can bump the CDN tagged versions in all the samples at the same time).

patrickarlt:
@jgravois let me do a release first and make sure everything is up to date. Then I want to switch to jsdelivr.com CDN where ever we can.

jgravois:
:+1:

PR #513, opened 2015-04-21
allow passing of transparent option

qbit:
I was pulling in some layers and I noticed I couldn't specify transparent in the options obj..


patrickarlt:
Could you also add a default value of `true` here? https://github.com/Esri/esri-leaflet/pull/513/files#diff-92a727b8d8cc44ec622b0a0335471d15L5 we cant pass `undefined` to the API.

qbit:
Added. I can squash it if needed as well.

patrickarlt:
I'm far to bad about squashing my commits to ask it of others :smile:

Merging.

qbit:
:+1:

PR #515, opened 2015-04-24
add error trap within esri.Tasks.IdentifyFeatures.run()

### LINKED ISSUE #512: CORS error is not handled properly when using identify
I'm doing an identify operation on a layer, which is failing due to cross origin policy. However, the error never bubbles back to the caller due to being masked by a secondary error:

image

image

jgravois:
closes #512

### DIFF
--- a/site/source/pages/examples/identifying-features.hbs
+++ b/site/source/pages/examples/identifying-features.hbs
@@ -36,15 +36,17 @@ layout: example.hbs
       pane.innerHTML = 'Loading';
     }
     federalLands.identify().on(map).at(e.latlng).run(function(error, featureCollection){
-      identifiedFeature = new L.GeoJSON(featureCollection.features[0], {
-        style: function(){
-          return {
-            color: '#5C7DB8',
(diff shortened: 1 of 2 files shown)


patrickarlt:
My brain is having some issues with how we process this data. The code flow just doesn't look that good and is pretty hard to understand.

What about something like this

```js
// immediately invoke with an error
if(error) {
  callback.call(context, error, undefined, response);
  return;

// ok no error lets just assume we have features...
} else {
  var featureCollection = EsriLeaflet.Util.responseToFeatureCollection(response);
  response.results = response.results.reverse();
  for (var i = 0; i < featureCollection.features.length; i++) {
    var feature = featureCollection.features[i];
    feature.layerId = response.results[i].layerId;
  }
  callback.call(context, undefined, featureCollection, response);
}
```

This feels a little more sane and easier to understand.

jgravois:
agreed and fixed.

also added a trap in our sample so that it doesn't throw errors when identify doesn't return a feature.

Issue #521, opened 2015-04-29
Don't replace double-quotes in where string


blkqi:
Currently EsriLeaflet.Tasks.Query will replace any and all double quotes (") in a where string with quoted single-quotes (\'). Double quotes are perfectly valid in SQL and some backends require them in order to properly specify certain identifiers.

patrickarlt:
I have long wondered if this would cause issues. The original reason this happens is that ArcGIS Online does not accept double quotes. Anyone attempting to do something like `layer.setWhere('BIKEMODE="Low traffic through street"')` would actually produce an error. Since the error message is totally unhelpful I decided to replace double quotes with single quotes to make it work.

**Double  quotes** (error) [http://services.arcgis.com/uCXeTVveQzP4IIcx/ArcGIS/rest/services/Bike_Routes/FeatureServer/0/query?where=BIKEMODE="Low traffic through street"](http://services.arcgis.com/uCXeTVveQzP4IIcx/ArcGIS/rest/services/Bike_Routes/FeatureServer/0/query?where=BIKEMODE="Low traffic through street")

**Single  quotes** (no error) [http://services.arcgis.com/uCXeTVveQzP4IIcx/ArcGIS/rest/services/Bike_Routes/FeatureServer/0/query?where=BIKEMODE='Low traffic through street'](http://services.arcgis.com/uCXeTVveQzP4IIcx/ArcGIS/rest/services/Bike_Routes/FeatureServer/0/query?where=BIKEMODE='Low traffic through street')

I'm to see if this is something that can be resolved in the backend of ArcGIS Online and then we can eliminate this.

patrickarlt:
@blkqi do you have example of a public facing service that I can use as an example besides something hosted/in Online?

patrickarlt:
By that I mean something where you absolutely MUST use doublequotes.

blkqi:
@patrickarlt the first form of your where clause produces an error because it is syntactically invalid. The second form is correct and so is the following form which utilizes double-quotes [1]: 
```
"BIKEMODE"='Low traffic through street'
```
See: 
[...]

PR #524, opened 2015-05-01
stop replacing double quotes in where strings

jgravois:
instead:
 * add an example and link to more SQL syntax help in the API reference
 * provide an error in the console explaining that doublequoted strings are a common reason for 400 errors
 * make existing tests syntactically correct

### DIFF
--- a/src/Tasks/Query.js
+++ b/src/Tasks/Query.js
@@ -67,8 +67,8 @@ EsriLeaflet.Tasks.Query = EsriLeaflet.Tasks.Task.extend({
   },
 
   where: function(string){
+    // instead of converting double-quotes to single quotes, pass as is, and provide a more informative message if a 400 is encountered
     this.params.where = string;
-    // string.replace(/"/g, "\'"); // jshint ignore:line
     return this;
   },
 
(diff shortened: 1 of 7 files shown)


src/Tasks/Query.js
```diff
@@ -97,14 +98,16 @@ EsriLeaflet.Tasks.Query = EsriLeaflet.Tasks.Task.extend({
...
-      }, context);
+      }, this);
 
     // otherwise convert it in the callback then pass it on
     } else {
       return this.request(function(error, response){
+        this._trapSQLerrors(error);
         callback.call(context, error, (response && EsriLeaflet.Util.responseToFeatureCollection(response)), response);
-      }, context);
+      }, this);
```
jgravois: context is `undefined` within this callback.  i needed to replaced with `this` so that i could call an internal method...

jgravois:
closes #521

patrickarlt:
@jgravois It looks like the tests get into some kind of infinite loop. Can you fix this an then I can merge this. https://travis-ci.org/Esri/esri-leaflet/builds/60904637

jgravois:
ok. in order to fix the tests i needed to interrogate errors a little more deeply to confirm they were actually 400s from the server (and not readyState errors from within `Request`.  all is well now.

PR #527, opened 2015-05-09
Base classes

### LINKED ISSUE #497: remove/demote L.esri.Services.FeatureLayer from API reference
i think the current placement of [L.esri.Services.FeatureLayer](http://esri.github.io/esri-leaflet/api-reference/services/feature-layer.html) in the API reference is unnecessarily confusing.  i find myself landing there time and time again when i really want to be reading up on what's available to `L.esri.Layers.FeatureLayer`

if we're going to include it, we should be more explicit in the note at the top and link back to `L.esri.Layers.FeatureLayer` 

alternatively, perhaps we could/should doc `FeatureGrid` and `FeatureManager` as well and make a point of explicitly segregating all three of them.

jgravois:
closes #497

rearranged class organization in API reference to group `Services`, utility methods and `Events` under the heading 'base classes'
this has the effect of promoting `L.esri.Layers.FeatureLayer` and other `Layers` as the classes that developers typically need to concern themselves with.

looking at the actual source code more, i no longer think it would be helpful to write API reference pages for `L.esri.Layers.FeatureGrid` and `L.esri.Layers.FeatureManager`.  @patrickarlt is right that one (or a few) conceptual articles explicitly describing **how** to extend classes would be a lot more useful.


patrickarlt:
@jgravois I like where this is goind but I think the optimal order might be

* Layers
* Tasks
* Misc. (Request/Util/Events)
* Base Classes (Services for now. I'll add conceptual/how to doc later)

patrickarlt:
I like the idea of keeping base classes and the Request/Util/Events stuff separated.

jgravois:
sounds great to me.  layed down one more commit to rearrange as suggested.
also took the opportunity to bump `L.esri.Layers.FeatureLayer` a little higher in the list order within `Layers`

patrickarlt:
@jgravois merging this. Will you run a deploy to get this in production.

jgravois:
done and done.  thx @patrickarlt!

PR #533, opened 2015-05-18
make sure appropriate icon for each marker is used when calling FeatureLayer.refresh();

jgravois:
#### problem:
currently we fail to restyle features after fetching from the server.  

because point features implement L.Marker icons, existing methods pertaining to setting `style` aren't relevant.

#### STR:
1. load sample app `debug/refresh-style` and use the button to update the status of the test feature so that it has an attribute which should alter the icon its drawn with.
2. refresh the feature layer (using included button) and notice that the symbology of the feature in the topleft corner has not changed.

#### changes introduced in this pull request:
* ensure that new attributes from the server are injected in clientside feature collections (and not just changes to geometry)
* renamed `FeatureLayer._updateLayerGeometry()` to `_updateLayer()` to reflect the new addition
* add a new check within `createLayers()` so that we appropriately trap changes to point feature geometries
* make sure a call is made to `setIcon` to refresh symbology appropriately.

#### to do:
- [x] still need to test with polylines to confirm that all that's necessary for comparable coverage is a call to `resetStyle()`

cc/ @kneemer.  

i'm curious to hear how you in particular feel about these changes.  especially in regard to how it will affect esri-leaflet-renderers

### DIFF
--- a/debug/refresh-style.html
+++ b/debug/refresh-style.html
@@ -73,7 +73,7 @@
   <div id="map"></div>
   <div id="info-pane" class="leaflet-bar">
     <label>
-    <button onclick='postEdits()'>update status attribute of objectid: 149478</button><br>
+    <button onclick='postEdits()'>update 'status' attribute of objectid: 149478</button><br>
     <button onclick='refresh()'>refresh feature layer</button>
     </label>
   </div>
@@ -102,7 +102,17 @@
     }).addTo(map);
 
     function postEdits () {
-      var updated = [{"geometry": { "x": -117.20879371400002, "y": 34.05838465599999 }, "attributes":{"OBJECTID":149478,"status":0 }}];
+      /*
+      original:
+      X: -117.20879371400002
+      Y: 34.05838465599999
+
+      freeway-ish
+      -117.2015
+      34.0635
+      */
+
+      var updated = [{"geometry": { "x": -117.20879371400002, "y": 34.05838465599999 }, "attributes":{"OBJECTID":149478,"status":bool }}];
 
       L.esri.post('http://sampleserver6.arcgisonline.com/arcgis/rest/services/RedlandsEmergencyVehicles/FeatureServer/0/updateFeatures', {
           features: updated,
@@ -112,7 +122,7 @@
           console.log(error);
         } else {
           console.log(response);
-          bool = 1;
+          bool = 1 - bool;
         }
       });
     }
--- a/debug/refresh-style.html
+++ b/debug/refresh-style.html
@@ -73,7 +73,11 @@
   <div id="map"></div>
   <div id="info-pane" class="leaflet-bar">
     <label>
-    <button onclick='postEdits()'>update 'status' attribute of objectid: 149478</button><br>
+    status '0' - blue icon<br>
+    status '1' - red icon<br><br>
+
+    <button onclick='postEdits()'>update 'status' attribute of objectid: 149478</button><br><br>
+    <div id='status-status'>current status of objectid: 149478 - ?</div><br>
     <button onclick='refresh()'>refresh feature layer</button>
     </label>
   </div>
@@ -123,6 +127,7 @@
         } else {
           console.log(response);
           bool = 1 - bool;
+          document.getElementById('status-status').innerHTML = 'current status of objectid: 149478 - ' + bool;
         }
       });
     }
--- a/debug/refresh-style.html
+++ b/debug/refresh-style.html
@@ -125,9 +125,9 @@
         if(error){
           console.log(error);
         } else {
-          console.log(response);
-          bool = 1 - bool;
+          // console.log(response);
           document.getElementById('status-status').innerHTML = 'current status of objectid: 149478 - ' + bool;
+          bool = 1 - bool;
         }
       });
     }
--- a/debug/refresh-style.html
+++ b/debug/refresh-style.html
@@ -125,7 +125,6 @@
         if(error){
           console.log(error);
         } else {
-          // console.log(response);
           document.getElementById('status-status').innerHTML = 'current status of objectid: 149478 - ' + bool;
           bool = 1 - bool;
         }
--- a/debug/refresh-style.html
+++ b/debug/refresh-style.html
@@ -1,0 +1,126 @@
+<!DOCTYPE html>
+  <html>
+  <head>
+    <meta charset=utf-8 />
+    <title>refresh point icons</title>
+    <meta name='viewport' content='initial-scale=1,maximum-scale=1,user-scalable=no' />
+
+    <!-- Load Leaflet from CDN-->
+    <link rel="stylesheet" href="http://cdn.leafletjs.com/leaflet-0.7.3/leaflet.css" />
+    <script src="http://cdn.leafletjs.com/leaflet-0.7.3/leaflet.js"></script>
+
+    <!-- Load Esri Leaflet from source-->
+    <script src="../src/EsriLeaflet.js"></script>
+    <script src="../src/Request.js"></script>
+    <script src="../src/Util.js"></script>
+    <script src="../src/Controls/Logo.js"></script>
+
+    <script src="../src/Services/Service.js"></script>
+    <script src="../src/Services/FeatureLayer.js"></script>
+    <script src="../src/Services/ImageService.js"></script>
+    <script src="../src/Services/MapService.js"></script>
+
+    <script src="../src/Layers/FeatureLayer/FeatureGrid.js"></script>
+    <script src="../src/Layers/FeatureLayer/FeatureManager.js"></script>
+    <script src="../src/Layers/FeatureLayer/FeatureLayer.js"></script>
+
+    <script src="../src/Layers/RasterLayer.js"></script>
+    <script src="../src/Layers/ImageMapLayer.js"></script>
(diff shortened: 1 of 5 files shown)


debug/refresh-style.html
```diff
@@ -0,0 +1,141 @@
...
+      L.esri.post('http://sampleserver6.arcgisonline.com/arcgis/rest/services/RedlandsEmergencyVehicles/FeatureServer/0/updateFeatures', {
+          features: updated,
+          f: 'json'
+        }, function(error, response){
+        if(error){
+          console.log(error);
+        } else {
+          console.log(response);
+          bool = 1 - bool;
+          document.getElementById('status-status').innerHTML = 'current status of objectid: 149478 - ' + bool;
```
kneemer: I know this is just for debugging, but shouldn't line 129 and 130 be swapped? I was getting confusing current status info.

jgravois: yes.  totally.

src/Layers/FeatureLayer/FeatureLayer.js
```diff
@@ -90,8 +103,20 @@ EsriLeaflet.Layers.FeatureLayer = EsriLeaflet.Layers.FeatureManager.extend({
...
+        // this.resetStyle();  ?
+        return;
+      }
+
+      // points
+      if (layer && layer.setLatLng) {
+        this._updateLayer(layer, geojson);
+        var iconTest = this.options.pointToLayer(geojson, L.latLng(geojson.geometry.coordinates[1], geojson.geometry.coordinates[0]));
+        var updatedIcon = iconTest.options.icon;
+        layer.setIcon(updatedIcon);
```
kneemer: The iconTest name is a little confusing here, are you expecting that there might not be an icon or a pointToLayer function?  If so, what happens in layer.setIcon if updateIcon is null?  Does it break or just use the default leaflet icon?

jgravois: it just breaks of course.

maybe this?

```javascript
if (this.options.pointToLayer){
  // original logic
}
else {
  // don't bother calling .setIcon()
}
```

kneemer: That is easier to follow.  Is it possible to have a pointToLayer function and no icon?  It doesn't seem easy to do, just wondering.

I guess what I'm wondering the most is what scenarios might occur that we should be setting the icon back to the default leaflet icon.

jgravois: from the [doc](http://leafletjs.com/reference.html#geojson-pointtolayer) the method seems to be specifically geared toward supplying custom symbology, not that that's a guarantee...

that being said, i feel reasonably comfortable with the current logic.

jgravois:
added a commit after confirming that a call to `resetStyle(id)` within `FeatureLayer.createLayers()` ensures that custom symbology is applied correctly to polylines (and doesn't create trouble w/ default styles).

src/Layers/FeatureLayer/FeatureLayer.js
```diff
@@ -90,8 +99,23 @@ EsriLeaflet.Layers.FeatureLayer = EsriLeaflet.Layers.FeatureManager.extend({
...
       if (layer && layer.setLatLngs) {
-        this._updateLayerGeometry(layer, geojson);
+        this._updateLayer(layer, geojson);
+        // update custom symbology for layer if necessary
+        this.resetStyle(geojson.id);
+        return;
+      }
+
+      // points
+      if (layer && layer.setLatLng) {
```
patrickarlt: This doesn't check for `layer.setIcon` so if a layer has `setLatLng` but not `layer.setIcon` (like `L.CircleMarker`) it will fail.

Probably best to do a separate check entirely like `if (layer && layer.setIcon)`.

jgravois:
```
if (layer && layer.setLatLng) {
  this._updateLayer(layer, geojson);
  if (layer && layer.setIcon) {
    // update custom symbology, if necessary
    if (this.options.pointToLayer){
      var getCustomIcon = this.options.pointToLayer(geojson, L.latLng(geojson.geometry.coordinates[1], geojson.geometry.coordinates[0]));
      var updatedIcon = getCustomIcon.options.icon;
      layer.setIcon(updatedIcon);
    }
  }
  return;
}
```
?

patrickarlt:
@jgravois how does this handle updating things like `L.circleMarker`? I want to make sure this works if you return a `L.circleMarker` as opposed to just a `L.Marker` from `pointToLayer`.

jgravois:
not sure, but i can confirm.

src/Layers/FeatureLayer/FeatureLayer.js
jgravois: nevermind.  i see what you're saying now.  i'll rewrite this after i confirm whats needed.

jgravois:
logic has been added to handle applying `L.circleMarker` styles appropriately.  

i noticed that `refresh()` is still not 100% rock solid.  after stepping through the code it looks like sometimes features from new queued gridded queries don't make it inside [createFeatures()](https://github.com/Esri/esri-leaflet/blob/master/src/Layers/FeatureLayer/FeatureLayer.js#L81).

you can see this in `refresh-circle.html`.  when zooming in tight on the top center feature, refresh is consistent.  when zoomed out to display all 6 features, sometimes the feature symbology doesn't change.

patrickarlt:
@jgravois can you fix these failing tests? I'll take a look at `refresh()` when I get a chance.

jgravois:
my bad.  those tests were failing because i had old code from #535 in there.  ive fixed and all tests are passing now.

jgravois:
added a corresponding test.  it wont pass until after #535 is merged.

jgravois:
threw in a test that confirms polyline symbology is instantiated appropriately (and that we restyle correctly on `refresh()`) for good measure.

this is ready to merge.

patrickarlt:
@jgravois merged your other PR and it looks like this is failing. Can you fix and I'll merge?

jgravois:
ok.  merged `master`, fixed the conflicts and removed my own inappropriate early `returns`.  all tests passing locally now.

src/Layers/FeatureLayer/FeatureLayer.js
```diff
@@ -89,8 +98,37 @@ EsriLeaflet.Layers.FeatureLayer = EsriLeaflet.Layers.FeatureManager.extend({
         this._map.addLayer(layer);
       }
 
+      // polylines and polygons
       if (layer && layer.setLatLngs) {
-        this._updateLayerGeometry(layer, geojson);
+        this._updateLayer(layer, geojson);
+        // update custom symbology, if necessary
+        this.resetStyle(geojson.id);
```
patrickarlt: Are you sure you want to `resetStyle ` here? This code just updates the geometry. If I had applied custom styling to an item (like on click) I would not expect it to get reset when I zoomed in.

src/Layers/FeatureLayer/FeatureLayer.js
```diff
@@ -89,8 +98,37 @@ EsriLeaflet.Layers.FeatureLayer = EsriLeaflet.Layers.FeatureManager.extend({
...
+        }
+        // L.circleMarker check
+        if (layer && layer.setStyle) {
+          if (this.options.pointToLayer){
+            var getStyle = this.options.pointToLayer(geojson, L.latLng(geojson.geometry.coordinates[1], geojson.geometry.coordinates[0]));
+            var updatedStyle = getStyle.options;
+            this.setFeatureStyle(geojson.id, updatedStyle);
+          }
+          else {
+            this.resetStyle(geojson.id);
```
patrickarlt: Same goes for here. I realize that is so we can catch updated rendering based on attributes but it also seems like it could be unwanted behavior.

patrickarlt:
This is turning into some seriously messy conditionals. I wonder if we can simplify this like so:

```js
// update geomerty if neccessary
if (layer && (layer.setLatLngs || layer.setLatLng)) {
  this._updateLayer(layer, geojson);
}

// if this looks like a marker
if (layer && layer.setIcon && this.options.pointToLayer) {
  // update custom symbology, if necessary
  if (this.options.pointToLayer){
    var getIcon = this.options.pointToLayer(geojson, L.latLng(geojson.geometry.coordinates[1], geojson.geometry.coordinates[0]));
    var updatedIcon = getIcon.options.icon;
    layer.setIcon(updatedIcon);
  }
}

// looks like a vector marker (circleMarker)
if (layer && layer.setStyle && this.options.pointToLayer) {
  var getStyle = this.options.pointToLayer(geojson, L.latLng(geojson.geometry.coordinates[1], geojson.geometry.coordinates[0]));
  var updatedStyle = getStyle.options;
  this.setFeatureStyle(geojson.id, updatedStyle);
}

// looks like a path (polygon/polyline)
if(layer && layer.setStyle && this.options.style) {
  this.resetStyle(geojson.id);
}

patrickarlt:
@jgravois I'm starting to wonder if we really need to handle style updates in `createLayer`. The whole point of this is to make sure styles update when attributes on the features change. So what if we move all this styling updating logic to `resetStyle` (thus making it handle `Marker`, `CircleMarker` and `Path`) or a new method like `redraw(id)` and then calling that method from inside `addFeature`?

jgravois:
this is great feedback.  i'll give the suggested refactor a go.

PR #535, opened 2015-05-19
add circleMarker check to createLayers

### LINKED ISSUE #534: L.CircleMarker symbols aren't instantiated with the options supplied in constructor
the problem in `L.esri.Layers.FeatureLayer` is evident in [this](http://esri.github.io/esri-leaflet/examples/querying-feature-layers-2.html) sample.

all of the `L.CircleMarker` constructor [options](http://leafletjs.com/reference.html#circlemarker) are being ignored.
```javascript
// commenting these options out or substituting different values doesn't affect marker symbology
{ 
  color: '#5B7CBA', 
  weight: 2,
  opacity: 0.85,
  fillOpacity: 0.5
}
```
`.setFeatureStyle()` is working as expected.

jgravois:
closes #534

added a check into `FeatureLayer.createLayers()` to ensure that a null style isn't passed accidently when a client application leverages `L.circleMarker` symbols for point features.

resolves a bug that i'm pretty sure was introduced by changes here 28f4d980222cccacd31e6a3b802d7d18436032d4

to be perfectly honest, my brain isn't working well enough to comprehend what other edge cases we need to make sure we are trapping, but i'll chew on it tomorrow.

### DIFF
--- a/spec/Layers/FeatureLayer/FeatureLayerSpec.js
+++ b/spec/Layers/FeatureLayer/FeatureLayerSpec.js
@@ -51,11 +51,25 @@ describe('L.esri.Layers.FeatureLayer', function () {
       }
   })];
 
+  var point = [({
+      type : 'Feature',
+      id: 1,
+      geometry: {
+        type: 'Point',
+        coordinates: [-95, 43]
+      },
+      properties: {
+        time: new Date('Febuary 1 2014').valueOf()
+      }
+  })];
+
   beforeEach(function(){
     layer = L.esri.featureLayer('http://gis.example.com/mock/arcgis/rest/services/MockService/MockFeatureServer/0', {
       timeField: 'time',
       pointToLayer: function(feature, latlng){
-        return L.circleMarker(latlng);
+        return L.circleMarker(latlng, {
+          color: 'green'
+        });
       }
     }).addTo(map);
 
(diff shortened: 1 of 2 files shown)


jgravois:
hold off on merging while i investigate potentially legitimately failing tests.

src/Layers/FeatureLayer/FeatureLayer.js
```diff
@@ -98,7 +98,16 @@ EsriLeaflet.Layers.FeatureLayer = EsriLeaflet.Layers.FeatureManager.extend({
...
         newLayer =  this.createNewLayer(geojson);
         newLayer.feature = geojson;
-        newLayer._originalStyle = this.options.style;
+
+        if (this.options.style) {
+          newLayer._originalStyle = this.options.style;
+        }
+
+        // circleMarker check
+        if (newLayer.setStyle) {
```
patrickarlt: This would really just trap anything with `setStyle` including `L.Polygon`, `L.Path`, ect... which is probably something we want to do anyway.

patrickarlt:
Since I'm doing a lot of this in the examples could you build this and test all the examples with it?

jgravois:
okay, i needed to make a correction to the logic.  all tests passing now (and examples all behave as expected)

patrickarlt:
@jgravois is this good to go?

jgravois:
yup.  i'll try and follow up soon with some test coverage.

jgravois:
added a corresponding test.  this is ready to merge.

spec/Layers/FeatureLayer/FeatureLayerSpec.js
```diff
@@ -230,6 +258,8 @@ describe('L.esri.Layers.FeatureLayer', function () {
 
     layer.createLayers(multiPolygon);
 
+    expect(layer.getFeature(1).getLayers()[0].options.color).to.equal('black');
```
patrickarlt: @jgravois this is only going to check that `setStyle`/`resetStyle` works on multipolygons. Could you add a specific test for `setStyle`/`resetStyle` when using `L.circleMarker`?

jgravois: i don't think we were having problems with setStyle/resetStyle before, but its a good test, so i followed your suggestion and laid it down in an additional commit

PR #550, opened 2015-05-29
Introduce new redraw() method for programmatically updating feature symbology

patrickarlt:
@jgravois I'm opening this up as a counter to https://github.com/Esri/esri-leaflet/pull/533.

My main problem with the work that you did in https://github.com/Esri/esri-leaflet/pull/533 is that it would override user defined styles whenever a feature was pulled from the server so for example imagine this case:

1. Load feature `1` apply default style
2. User clicks feature `1`, style changes from default to selected
3. User zoom in triggering a reload of feature `1` from the server
4. Selected styles are wiped out

However this is also pretty undesirable:

1. Load feature `1` apply default style
2. Feature `1` changes in the service (due to an edit)
3. Feature `1` rendering stays the same and should be different

I really want to avoid unexpectedly clobbering styles that a developer defines. This simply introduces a new `redraw(id)` method which allows us to cover a few different use cases.

##### 1. Refresh all features every X seconds

```js
setTimeout(function(){
 featureLayer.refresh(); // redraw() is called automatically after the next load event
}, 1000);
```

##### 2. Update feature symbology after an update

```js
featureLayer.updateFeature(geojson, function(error, response){
  featureLayer.redraw(geojson.id);
});
```

##### 3. Update feature symbology after every load

```js
featureLayer.on('load', function(){
  featureLayer.eachFeature(function(layer){
    layer.redraw(layer.feature.id);
  }
}
```


jgravois:
you dog you :smile:
I'll take a deeper look *really* soon.

jgravois:
this looks really good.  i have a suggestion though.  ultimately i think we want to make a method available to refetch features without restyling and another to fetch and restyle (and still avoid clobbering custom clientside rendering if possible).

```javascript
fl.on('click', function(evt) {
  fl.setFeatureStyle(evt.layer.feature.id, {
    color: 'purple'
  });
});
```
after adding the listener above to our debugging sample app, i confirmed that calling `redraw()` still overrides custom styles defined on the clientside.  this got me thinking that we can be much more selective in our restyling by taking the opportunity afforded by [_updateLayer()](https://github.com/patrickarlt/esri-leaflet/blob/refresh/src/Layers/FeatureLayer/FeatureLayer.js#L58) to keep tabs on whether a feature's attributes are really any different and only restyle if we noticed a change.

patrickarlt:
You are still explicitly calling `redraw()` in the case above though which is fine since you KNOW what is going on. You called `redraw()` and you should expect your feature to be redrawn with its original style.

Property checking seems a little dirty and confusing to me. It might get us into weird spots like "Only redraw if an attribute changed" or "Don't redraw if the style is different from the original style" which is harder to explain since it restyles in some cases but not others. This would also break a whole bunch of use cases for adding state in `layer.features.properties`, like for tracking the state of a layer (editing, edited, hovering, clicking) on the client.

I'm thinking maybe the following would work:

* `redraw(id)` - redraw a specific feature with its original style
* `redraw()` - redraw all features in their original style
* `refresh(id, callback)` - fetch feature from server and update its properties
* `refresh(callback)` - refresh all features from server and update local properties

I think rethinking the scope of `refresh` is probably outside this issues though. I would also like to play around with `updateFeature()` updating things in place rather then destroying and recreating everything but that is also a separate PR.

jgravois:
that all sounds good to me.  glad to see you merged and tagged RC8.  

sorry i've been slow to respond.  i've been on a 7 day out of towner..

PR #557, opened 2015-06-05
add livereload support for those building the website locally

### LINKED ISSUE #556: need to update doc to clarify that bounds() requires ArcGIS Server 10.3.1
although the doc [here](http://resources.arcgis.com/en/help/arcgis-rest-api/index.html#/Query_Feature_Service_Layer/02r3000000r1000000/) states that `returnExtentOnly` was added to ArcGIS Server queries at 10.3, it actually wasn't shipped until 10.3.1.

jgravois:
i thought it would be cool if we added [livereload](https://github.com/gruntjs/grunt-contrib-watch#optionslivereload) support for those that build the website.  it was pretty straightforward excepting the fact that i don't know how to dynamically insert script tags into built pages in development only.

is using something like [grunt-script-link-tags](https://github.com/andrewjmead/grunt-script-link-tags) and ensuring that the task isn't called when building for production a good approach?  

@ngoldman or @paulcpederson your feedback would be sincerely appreciated.

unrelated fix closes #556 as well.


patrickarlt:
FWIW [Browsersync](http://www.browsersync.io/) is my go-to livereload solution and in general the only one I can ever get to work.

paulcpederson:
@jgravois I think the way I'd approach not adding the script to the page on build is to do it via assemble. Right now it looks like there are two different targets being run by grunt-assemble: `dev` and `build`. Assuming those are the two builds and you only want the script tag in `dev`, I think you might be able to add a flag as `data` and then just use a conditional in the template. 

I've never used assemble, but the docs for their grunt task seem to indicate you can set data as a simple object and then use it in templates. Example:

```
<!-- inside partial -->
{{#if isDev}}
  <script src="//localhost:35729/livereload.js"></script>
{{/if}}
```

And then in the assemble Grunt task:

```
      dev: {
        options: {
          data: {
            isDev: true
          },
          assets: 'site/build/'
        }
      }
```

I have no idea whether that would work or not, but if it did it would keep you from having to add another grunt plugin into the mix.

paulcpederson:
I will also say that lately I have been using [live-server](https://www.npmjs.com/package/live-server) and it seems to work ok. It doesn't require adding a script tag to the document because it just adds the script tag itself when it serves the file. I don't know how well that would play with the current tooling, though.

jgravois:
i couldn't get a simple data object working, but it everything was hunkydory after i added it in an external .json file.

thanks for the help @paulcpederson!  this is ready for a merge.

PR #579, opened 2015-07-08
new zoom to bounds sample for legacy feature services

jgravois:
* added new sample to demonstrate how to zoom to the extent of features returned by ArcGIS Server versions that don't support 'bounds' directly
* fixed typo in header
* updated hyperlink in FeatureLayer sample header info to use a relative path.  existing url was invalid locally because 'esri-leaflet' isn't present.


jgravois:
in my own testing (at `rc.8`), the zooming in this sample is a bit finicky, particularly when features are already present in the browser cache.

that being said, everything looks good with `master` so i'm just going to double-check to make sure all is well when we tag again.

Issue #578, opened 2015-07-08
Switch to passing URL in options for layers?


patrickarlt:
In RC 4 I refactored all `Task` and `Service` objects to accept `url` in the options like so:

```js
L.esri.Tasks.query({
  url: 'https://...'
});
```

I've been debating making the change for all the layers for a few reasons.

1. Everything now accepts `options` as its first argument, except maybe `BasemapLayer`
2. We can warn about not passing `url` more efficiently which we currently don't do
3. Opens the door for things like https://github.com/Esri/esri-leaflet/issues/570 later
4. Makes passing params to the underlying services on layers easier

The major problem is that if this goes into 1.0 we have 2 different patterns for layer initialization across the 2 branches 1.0.0 vs 0.9.0. I might be able to make this backward compatible by checking if the first parameter is an object though.

@jgravois @ngoldman

jgravois:
my gut reaction is that we're better off tackling the refactor now (rather than later) and that we can probably get away with skipping effort on backwards compatibility until such a time that we come across a bug that we can't conveniently port back to 0.9.x.

that being said, i don't have any plans personally to tackle #570 and i haven't encountered any problems with the current architecture so i'm pretty ambivalent about the whole thing.

if you want to go ahead with it, i think it'd be worthwhile to package a snapshot of the documentation site that corresponds with 0.9.x and either host it ourselves or at a bare minimum make it available for download without requiring a build.

ngoldman:
I would recommend normalizing towards an options object across public functions for the sake of both releases (for Leaflet 0.7.3 and 1.0.0), as it's a better pattern and a stable release with a major version change is the time to do that. This would follow with my recommendation re: #577.

@patrickarlt @jgravois

PR #581, opened 2015-07-09
Constructor updates

patrickarlt:
* Implements `url` in `options` https://github.com/Esri/esri-leaflet/issues/578
* Don't document `new Thing()` for object constructors https://github.com/Leaflet/Leaflet/issues/3408#issuecomment-100196640 encourage factories `thing()`
* Rename `L.esri.Services.FeatureLayer` to `L.esri.Services.FeatureLayerService`


CHANGELOG.md
```diff
@@ -1,8 +1,13 @@
 # Changelog
 
-## 0.9.0
+## 1.0.0
 
-This represents the stable release of Esri Leaflet that will be compatible with Leaflet 0.7.3. All future 0.9.X releases will be compatible with Leaflet 0.7.3 and contain only bug fixes.
```
jgravois: need to get rid of the 0.9.x reference within the text.

patrickarlt: red = removed. It all reads 1.0.0 now. But I need to update package.json

jgravois: wow.  please ignore me.

jgravois:
this looks really good!

what do you think about `L.esri.Services.FeatureService` instead of `L.esri.Services.FeatureLayerService`?  the second is a bit of a mouthful.

jgravois:
so stoked you figured out the issue with the timeouts in the DynamicMapLayer test!  i noticed our current string ( 'http://placehold.it/500x500&text=Image1' ) returns an image with no height.

looks like we should use 'http://placehold.it/500**x500**&text=Image1' instead.

patrickarlt:
I worry that `FeatureService ` could be confused with the actual [Feature Service](http://resources.arcgis.com/en/help/arcgis-rest-api/#/Feature_Service/02r3000000z2000000/) endpoint.

jgravois:
oh well.  i think `FeatureServiceLayer` might have made more sense, but its no big thing.

ngoldman:
agree with @jgravois's last comment if I'm understanding things correctly. It's a layer for feature services, not a service for feature layers, right?

Issue #575, opened 2015-07-07
metadata requests with an invalid token don't trigger the callback


jgravois:
```js
L.esri.Services.service({
  url: 'http://logistics.arcgis.com/arcgis/rest/services/World/ServiceAreas/GPServer/GenerateServiceAreas',
  token: 'badtoken'
}).metadata(function(error, response){
  // never triggered
  console.log(error);
});
```
this is because we currently:
 1. trap authentication errors
 2. push into `_requestQueue`
 3. intentionally don't call the callback.

(code [here](https://github.com/Esri/esri-leaflet/blob/master/src/Services/Service.js#L73-L80))

i don't know enough about the other use case we are trying to support to know if/how we can best 'fix' this problem.  i was just surprised to see the error swallowed while working on jgravois/esri-leaflet-gp/issues/17

jgravois:
substituting this would work (and i believe not screw up the current intention of trapping requests with no token at all).
```js
if (error && (error.code === 499 || error.code === 498 && !params.token)) {
```

patrickarlt:
This is actually how I intended it to work to avoid callbacks being called twice. Generally the expectation is that callbacks will only be called once. If we wanted to move to calling callbacks multiple times everyone would have to write their callbacks in such a way that they never ran code on errors. This means that all callbacks would have to be written like this:

```js
function(error, response, raw){
  if(error) {
    //handle error
  } else { 
   // handle success
  }
}
```

or

```js
function(error, response, raw){
  if(error) {
    //handle error
    return;
  }

  // handle success
}
```

because this means that if you write callbacks like this:

```
function(error, response, raw){
  if(error) {
    //handle error
  }

  // handle success
}

you will actually handle the success twice which is bad. This also means that if you DON'T handle errors your success callback will be called twice but only once with the actual success.

The other reason is that in cases like what you are doing there is no way to set a good token on the service.

```
L.esri.Services.service({
  url: 'http://logistics.arcgis.com/arcgis/rest/services/World/ServiceAreas/GPServer/GenerateServiceAreas',
  token: 'badtoken'
}).metadata(function(error, response){
  // so how do you retry with correct auth?
  console.log(error);
});
```

patrickarlt:
@jgravois we could handle things like this.

```js
L.esri.Services.service({
  url: 'http://logistics.arcgis.com/arcgis/rest/services/World/ServiceAreas/GPServer/GenerateServiceAreas',
  token: 'badtoken'
[...]

PR #584, opened 2015-07-09
allow handing auth errors in a callback

patrickarlt:
This allows for handling authentication errors in callbacks fixing https://github.com/Esri/esri-leaflet/issues/575.


jgravois:
i was **so** hopeful that `timeout` was going to solve our test problems.  everything else looks good!

patrickarlt:
@jgravois I'm not sure what it is. I think I might try to fix it by just creating some of our own images and hosting them locally.

PR #594, opened 2015-07-16
Add missing pipe

robertd:


jgravois:
good catch @robertd.  thanks!

PR #598, opened 2015-07-19
Esri Leaflet 2.0.0 (Leaflet 1.0 compat)

patrickarlt:


jgravois:
the aggregate file change is a net negative, just like Leaflet itself! :sparkling_heart:

PR #600, opened 2015-07-20
various sample fixes

jgravois:
went through and made a few fixes to the samples.  most of which are just spikes in the `z-index` value for info panes.

tbh, i'm not really sure exactly *why* L.esri.featureLayer.bindPopup() doesn't return a GeoJSON 'feature' as the callback function argument anymore.
 
tile-layer-2.html *should* work with `detectRetina` set to `true` on `L.esri.tiledMapLayer` but it doesn't, so i figured probably best to set as false until we figure out why.


jgravois:
layed down a few more commits (closes #602, #605)

PR #601, opened 2015-07-20
add maptiks script tag

jgravois:
@phpmaps wants to test out some [maptiks](http://get.maptiks.com/maptiks-v1/) map analytics.

if we want to keep this after the leaflet 1.0.0 doc update, i'll PR against the actual source.

we'll be able to check usage [here](https://analytics.maptiks.com/#/signin/)

### DIFF
--- a/examples/index.html
+++ b/examples/index.html
@@ -7,13 +7,13 @@
 <head>
   <meta charset="utf-8">
 
-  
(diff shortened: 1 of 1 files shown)


phpmaps:
Nice thanks @jgravois

phpmaps:
@patrickarlt this is to evaluate some of the analytics which maptiks generates

PR #603, opened 2015-07-24
fix script url

thinking-aloud:
The script URL was broken and did not work.


jgravois:
because of this jsdelivr/jsdelivr/issues/5796 'latest' returns the wrong version of our API.  for the time being, it'd probably be best for us to replace with:
```html
<script src="http://cdn.jsdelivr.net/leaflet.esri/1.0.0/esri-leaflet.js"></script>
```
if you're up for laying down another commit in this PR, i'd be happy to merge.

patrickarlt:
@jgravois now that I have managed to eliminate using `latest` from the web samples I think it would be best to just start hardcoding this and changing it every few releases.

jgravois:
agreed.

jgravois:
closed (via 63f735e17aa74c051535a662036c395ab6bec450)

PR #607, opened 2015-07-27
make sure package.json main tag references built file

### LINKED ISSUE #604: "main" field in package.json points to non-existent file (re v2.0.0-beta.4 published on npm)
`"main"` field in `package.json` points to "src/EsriLeaflet.js" which is not present in the published package, hence  `require('esri-leaflet')` doesn't work as expected.

Steps to reproduce:
```sh
npm i esri-leaflet@2.0.0-beta.4
cat node_modules/esri-leaflet/package.json | grep main # ..."src/EsriLeaflet.js"
ls node_modules/esri-leaflet/ # no 'src' folder
```

jgravois:
closes #604 (at least i'm 99% sure this is the best way)

### DIFF
--- a/package.json
+++ b/package.json
@@ -73,7 +73,7 @@
     "mapping"
   ],
   "license": "Apache-2.0",
-  "main": "src/EsriLeaflet.js",
+  "main": "dist/esri-leaflet.js",
   "readmeFilename": "README.md",
   "repository": {
     "type": "git",


jgravois:
changed my mind.  when you `npm install leaflet`, both the minified and raw source are laid down, along with the test suite, which seems useful.

patrickarlt:
@jgravois I'm actually going to revert this. The main reason is that browserify will actually use Babel on the fly to compile this directly from source. https://github.com/Esri/esri-leaflet/blob/master/package.json#L16-L21

patrickarlt:
The real fix is to add make `files` in package.json `['src', `dist`]`

patrickarlt:
I've got this in for the next version.

PR #617, opened 2015-08-18
make sure logo href opens in a new tab and TiledMapLayer regex fix

jgravois:
resolves #615 (despite the opinions of stalwarts like @swingley)

i noticed that the changes introduced in #614 were breaking tests (because `arcgisonline` wasn't defined and we were missing a closing parenthesis) so i included a fix for that as well.

lastly, i decided to use the small logo all the time to match the prominence of [Google's](https://developers.google.com/maps/documentation/javascript/) in version 3 of their JS API and to link to www.esri.com instead of the developers site.

Issue #370, opened 2014-10-13
Breakout website into its own repo


patrickarlt:
The Esri Leaflet website shouldn't really be tied to this repo. This means that issues with plugin docs have to end up here which seems a little confusing.

Moving the website also gives us the opportunity to update the design and incorperate docs for plugins.

jgravois:
not sure if you were planning on incorporating a redesign at the same time, but i broke out the website into its own repo [here](https://github.com/jgravois/esri-leaflet-website)

jgravois:
reminder of goal to include doc for plugins alongside the rest of the api.

jgravois:
now that we finally have the plugin doc in the esri leaflet website, i refreshed my previous work of breaking it out into its own seperate repo.  if you still want to do this, i'll ask the admins to move it into the [esri](http://github.com/Esri) organization after everyone's happy.

https://github.com/jgravois/esri-leaflet-doc
http://jgravois.github.io/esri-leaflet-doc/

should be dead simple.  the only two pending issues i can think of are:
1. creating an advanced configuration which loads `dist/esri-leaflet.js` from an entirely seperate directory (so that devs can test the samples against a custom build)
2. making sure the new site/url, redirects back (or vice versa)

PR #618, opened 2015-08-18
fix broken link for feature layer service

sidewalkballet:
I see the site has been broken out into https://github.com/jgravois/esri-leaflet-website - the 'edit this page' still links to esri-leaflet - not sure if this is the right place. 

cc @jgravois


jgravois:
thanks for catching that!  #370 is still a proposal (and probably should die)

tests aren't passing here because of #617.  all is actually well.

sidewalkballet:
:dancers:

PR #632, opened 2015-09-09
delete featurelayer._map

### LINKED ISSUE #609: Feature layer fires an error when removed from map (re 2.0.0-beta.4)
Specifically, `Uncaught TypeError: Cannot read property 'getBounds' of null` happening [here](https://github.com/Esri/esri-leaflet/blob/master/src/Layers/FeatureLayer/FeatureLayer.js#L171)

Reproduce:
1. Set up map
2. Add a FeatureLayer
3. Remove this FeatureLayer.

[fiddle](http://jsfiddle.net/64pympm0/) (see console)

Unfortunatelly I still didn't get a good idea of how the thing works to submit a descent PR, sorry.

jgravois:
i was still seeing some flaky behavior (ie: code was intermittently stepping into the block where we check for the presence of this._map) when removing feature layers so i dropped in an actual `delete` and wrote a test which confirms that the property is no longer set.

closes #609 

i'd like to tag/bump to 2.0.0 and drop the betas altogether right before i merge the new website changes in.

### DIFF
--- a/src/Layers/FeatureLayer/FeatureGrid.js
+++ b/src/Layers/FeatureLayer/FeatureGrid.js
@@ -21,6 +21,7 @@ export var FeatureGrid = L.Layer.extend({
 
   onRemove: function () {
     this._map.removeEventListener(this.getEvents(), this);
+    delete this._map;
     this._removeCells();
   },
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,11 +3,12 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
-[Upcoming changes][unreleased]
+### [Upcoming changes][unreleased]
 
 ### Fixed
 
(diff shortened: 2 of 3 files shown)


src/Layers/FeatureLayer/FeatureGrid.js
```diff
@@ -21,6 +21,7 @@ export var FeatureGrid = L.Layer.extend({
 
   onRemove: function () {
     this._map.removeEventListener(this.getEvents(), this);
+    delete this._map;
     this._removeCells();
```
kneemer: I'm curious what, this._removeCells() gives you if you are removing the map listeners and then the map?

jgravois: that's what triggers [removing](https://github.com/Esri/esri-leaflet/blob/master/src/Layers/FeatureLayer/FeatureLayer.js#L163-L192) the actual geojson features from the featureLayer (and induces errors when the map is still attached)

PR #638, opened 2015-09-14
New ArcGIS Portal API addition

phpmaps:
A proposed portal API base for the esri-leaflet maintainers to enjoy.  Please give it a critical read... I suspect there could be more conformance in the way the esri-leaflet API already handles request options and authorization, but I think what is shared here should be pretty close. 

The below sample that can be added to the build/examples/<portal> folder (it uses OAuth so remember the redirect_uri and app_id)

https://gist.github.com/phpmaps/ece4fabdf9d1863d36e9 
```

/*Initialize portal with endpoint and token */

portal = new L.esri.Portal({
  'url':"https://www.arcgis.com",
  'token':accessToken,
  'callback': function(error, result){
    console.log(result);
 }
});

/*Event fires when portal loads*/
portal.on('portal-loaded', function(evt){
  portal.query({
    'token': accessToken,
    'where': 'owner:doogle',
    'callback': function(error, result){
      console.log(result);
    }
 });
});

```

### DIFF
--- a/src/EsriLeaflet.js
+++ b/src/EsriLeaflet.js
@@ -13,6 +13,9 @@ export { Identify, identify } from './Tasks/Identify';
 export { IdentifyFeatures, identifyFeatures } from './Tasks/IdentifyFeatures';
 export { IdentifyImage, identifyImage } from './Tasks/IdentifyImage';
 
+// export portal
+export { Portal, portal } from './Portal/Portal';
+
 // export services
 export { Service, service } from './Services/Service';
 export { MapService, mapService } from './Services/MapService';
--- a/src/Portal/Portal.js
+++ b/src/Portal/Portal.js
@@ -1,0 +1,92 @@
+import { Service } from '../Services/Service';
+import { cleanUrl } from '../Util';
+
+export var Portal = Service.extend({
+	portalUrls: {},
+	  options: {
+	    url: "https://www.arcgis.com",
+	    f: 'json'
+	  },
+	  on: function(type, handler){
+	  		switch (type) {
+	      case 'portal-loaded':
+	      		L.Evented.prototype.on.call(this, 'portal-loaded', handler);
+	        break; 
+	      default: 
+	        return;
+	  		}
+	  },
+	  initialize: function(options) {
+	  	  var portal = this;
+	    var token = options.token ? options.token : null;
+	    if (!token) return console.log("Portal::initalize is missing token");
+	    this.options.url = options.url ? cleanUrl(options.url) : options.url;
+	    if ((options.proxy || options.token) && options.f !== 'json') {
+	      options.f = 'json';
+	    }
+	    L.Util.setOptions(this, options);
+	    return this.self({
+	      'token': token
+	    }, function(error, response) {
+	      if (options.callback) {
+	        options.callback.call(this, error, response);
+	        if (response) {
+	        	  portal.updatePortalPaths(response);
+	          L.Evented.prototype.fire.call(portal, 'portal-loaded');
+	        }
+	      }
+	    });
+	  },
+	  updatePortalPaths: function(self) {
+	  	  this.portalUrls.sharing = '/sharing/rest/';
+	    this.portalUrls.portals = '/sharing/rest/portals/';
+	    this.portalUrls.portal = '/sharing/rest/portals/' + self.id + '/';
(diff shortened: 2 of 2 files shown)


jgravois:
very cool @phpmaps!

i think it makes the most sense to release this functionality as a plugin (similar to our work on [renderers](https://github.com/esri/esri-leaflet-renderers), [GP](https://github.com/jgravois/esri-leaflet-gp), [related-tables](https://github.com/jgravois/esri-leaflet-related) etc.)

phpmaps:
@jgravois - Would you share more?  I've seen a few cases where modular code bases although good for some things get lost in the woodwork. This pull is the tip of the iceberg for bringing in all the different kinds of data sources and functionality for finding, saving and persisting content with the API.   If we want to incubate this API somewhere for tweaking and building out a few more features I am game - but ideally I think the Portal API could be core here. Typically I like the one and done experience which includes bringing in the projects you mentioned.  Maybe I am old school.  However the less ```<script src="another-file.js"></script>``` the better.  I don't want to get philosophical on you but think about it - it's the Portal  *ie. mother ship*.

jgravois:
> modular code bases...

i'm not issuing a decree here, but the Leaflet ecosystem is intentionally designed around isolating code that does one thing well into individual modules.  that's why my immediate reaction is to seperate the functionality.

> the less `<script src="another-file.js"></script>` the better.

totally agree.  i've been working on a browserify sample to share in a blog which demonstrates how easy it is to create a `bundle.js` file which wraps Leaflet, Esri Leaflet and Esri Leaflet Geocoder for this very reason.

> lost in the woodwork...

again, totally agree.  I still haven't merged them, but @patrickarlt and i made some changes to the Esri Leaflet website recently to integrate the API reference for plugins and showcase link to their own repositories for this very reason.

i promise i am **not** stonewalling you here Doug E. Doug.  just pointing out my initial opinion.  I'm happy to get involved with your work wherever it lives and it'd be trivial down the line to integrate it in core if we decided it was appropriate. :smile:

patrickarlt:
The problem with bundling this IS that Portal is the mothership. The amount of code that might ship in something like `L.esri.Portal` is probably equivalent to the entirety of Esri Leaflet itself and not everyone needs it. My general rule of thumb for things like this is:

1. If it has extra dependencies it goes in a plugin (cluster layer, vector basemaps, ect...)
2. If it isn't needed in (almost) every app then it goes in a plugin (webmap parsing, renderers, gp)
3. If it will add a significant amount of code or functionality it goes in a plugin (renderers, vector basemaps)
4. If it might need to be shared among plugins it goes in Esri Leaflet (core utils ect...)

I'm not saying that `L.esri.Portal` isn't something that people don't need, but my gut tells me to separate it out into a separate plugin.

ajturner:
This will be very useful @phpmaps 

I agree with @jgravois - Esri-leaflet is for displaying data, and has no dependency on Portal. Similarly A developer will want to work with Portal API but not have a map.

So this would be most useful as a standalone library that has a leaflet wrapper for loading into Esri-leaflet

tomwayson:
@phpmaps, I can see how a portal library could be used by all kinds of JS apps - not just those that use Leaflet. I agree w/ others above that such a library should at the very least be a plugin, and probably shouldn't have any dependency on Leaflet or Esri Leaflet at all. In fact, it might not to depend on anything other than [xhr](https://www.npmjs.com/package/xhr).

As far as modularity in general, as an app developer, I strongly prefer many small libraries. Package managers and modern module loaders like browserify, webpack, etc make adding, managing, and bundling dependencies much easier than script load order + usemin. @jgravois looking forward to that browserify example!

ajturner:
For xhr I recommend checking out [window.fetch polyfill](https://github.com/github/fetch)

phpmaps:
Thanks all for chiming into this discussion.  Certainly it felt good getting a portal class written and out there. @jgravois I look forward to the browserfy sample.  Once moved I hope we get some traction building out this *mother ship*. In putting together the loaded event (which took some head scratching) @mourner topic here was really helpful - https://github.com/Leaflet/Leaflet/issues/2280. There is now a little more implementation stuff that expands on that in the Portal class.

@gyllen You may want to check this out.  I will tag you where ever this project ends up.

tomwayson:
@ajturner the window.fetch polyfill looks nice, esp since it's API returns promises. Thanks!

PR #654, opened 2015-10-15
add some automatic watch/build/serve goodness

### LINKED ISSUE #650: start 'watching' raw source
we should incorporate a build tool which observes the source code and rebuilds the minified/concatenated `dist` folder content automatically.

jgravois:
resolves #650 

added a new script `npm run serve` which:
  * spins up a local webserver
  * automatically rebuilds when a change is made inside the `src` directory
  * launches `debug/sample.html` automagically

if anyone has any better ideas than this approach, they'd be warmly received.  whatever we end up shipping, i can lay down another commit later to make sure the README includes documentation.

cc/ @paulcpederson

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,10 +29,10 @@
     ]
   },
   "devDependencies": {
-    "eslint": "^0.24.1",
     "assemble": "^0.4.37",
     "babelify": "^6.1.3",
     "chai": "2.3.0",
+    "eslint": "^0.24.1",
     "esperanto": "^0.7.3",
     "gh-release": "^2.0.0",
     "grunt": "^0.4.2",
@@ -44,6 +44,7 @@
     "grunt-newer": "^0.7.0",
     "grunt-sass": "^1.0.0",
     "highlight.js": "^8.0.0",
+    "http-server": "^0.8.5",
     "isparta": "^3.0.3",
     "istanbul": "gotwarlost/istanbul.git#source-map",
     "karma": "^0.12.16",
@@ -55,6 +56,7 @@
     "karma-sourcemap-loader": "^0.3.5",
     "load-grunt-tasks": "^0.4.0",
     "mkdirp": "^0.5.1",
+    "nodemon": "^1.7.2",
     "phantomjs": "^1.9.17",
     "rollup": "^0.11.4",
     "semistandard": "^6.1.2",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -91,6 +95,11 @@
     "pretest": "npm run build",
     "test": "npm run lint && karma start",
     "release": "./scripts/release.sh",
-    "start": "grunt"
+    "serve": "./scripts/open.js && parallelshell 'npm run watch' 'http-server -p 5678'",
```
patrickarlt: You can actually make this a lot simpler. `npm run watch & http-server -p 5678 -c-1 -o`

`-c-1` turns off caching. `-o` tells `http-server` to just open up the root of the web server which I think is fine I would rather have less dependencies.

You also don't need `parallelshell`. If you need to run multiple commands at once. Just separate them with `&` it works on both Mac and Windows.

package.json
```diff
@@ -55,6 +56,9 @@
     "karma-sourcemap-loader": "^0.3.5",
     "load-grunt-tasks": "^0.4.0",
     "mkdirp": "^0.5.1",
+    "npm-watch": "^0.0.1",
```
patrickarlt: `npm-watch` is really just a thin wrapper around [nodemon](https://github.com/remy/nodemon) I would rather replace it with the actual nodemon command which would be `nodemon --watch src --exec "npm run build"`

This would mean that the final `serve` command would look like this:

```bash
nodemon --watch src --exec "npm run build" & http-server -p 5678 -c-1 -o
```

and you can get rid of the dependencies on `npm-watch`, `open`, and `parallelshell`

jgravois:
thx for the feedback @patrickarlt.  i had a feeling my initial proposal was a bit dependency heavy and i appreciate the help cleaning it up.  i've incorporated your feedback in a new commit and squashed everything down.

i'll be happy to see #655 resolved after this gets merged.

package.json
jgravois: not sure what i was doing wrong, but for some reason in my initial testing *watch* and *http-server* both needed their own shell.  perhaps because i was using a double ampersand?  either way, happy to be able to clean this up.

patrickarlt: the watcher and the server both need their own shell because they block subsequent commands from finishing so using `&&` doesn't work but `&` does. Here is a handy reference:

* `thing-1 && thing-2` - run `thing-1` then run `thing-2`
* `thing-1 & thing-2` - run `thing-1` and `thing-2` at the same time and combine the output

patrickarlt:
@jgravois http://blog.keithcirkel.co.uk/how-to-use-npm-as-a-build-tool/ is a really good reference.

package.json
jgravois: > both need their own shell because they block subsequent commands 

word. thanks for taking the time to explain!

PR #679, opened 2015-11-09
Use HTTPS on AJAX content (dist)

timwis:
The page currently doesn't render the map when viewed over HTTPS because it's pulling insecure content over HTTP

### DIFF
--- a/examples/styling-feature-layer-polygons.html
+++ b/examples/styling-feature-layer-polygons.html
@@ -178,7 +178,7 @@
   L.esri.basemapLayer('GrayLabels').addTo(map);
 
   L.esri.featureLayer({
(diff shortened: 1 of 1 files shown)


jgravois:
thanks for this @timwis!

would you be willing to lay down one more commit with the same change in our templated [source](https://github.com/Esri/esri-leaflet/blob/master/site/source/pages/examples/styling-feature-layer-polygons.hbs#L17)?  this will ensure your change isn't overwritten next time we push the entire site to `gh-pages`

timwis:
Sure thing -- can't do the same PR since it's a different branch so I made a new one

jgravois:
> it's a different branch

oh yeah.  duh.  my west coast brain is extra spacey.

PR #680, opened 2015-11-09
Use HTTPS on AJAX content (src)

timwis:
Revision of #679 
The page currently doesn't render the map when viewed over HTTPS because it's pulling insecure content over HTTP


jgravois:
thanks again @timwis!

PR #689, opened 2015-12-05
Break out grid dependency (take 2)

jgravois:


jgravois:
all green.  @patrickarlt if you have time, point me at the arcgis > geojson stuff you ported out and i can PR comparable updates here.

PR #695, opened 2015-12-17
ensure FeatureLayer max/minZoom are always respected

### LINKED ISSUE #643: Feature Layer maxZoom not respected when zooming out

jgravois:
resolves #643 (in favor of #660), based on direction provided by @patrickarlt [here](https://github.com/Esri/esri-leaflet/pull/660#issuecomment-151680624).  the tests are still pretty crap for this, but they're a little better than they were before.

testing manually everything looked good to me when:
* no min/maxZoom are declared in the constructor
* both are declared
* each are provided singularly
* when panning/zooming the map manually
* when calling methods like `map.setZoom()`
* when the initial map zoom level should draw the layer
* when it shouldn't

i'm not positive, but this might resolve the problem reported in JRHutson/Food-Resource-Map#37 as well.

### DIFF
--- a/spec/Layers/FeatureLayer/FeatureManagerSpec.js
+++ b/spec/Layers/FeatureLayer/FeatureManagerSpec.js
@@ -815,7 +815,7 @@ describe('L.esri.FeatureManager', function () {
     expect(requestendSpy.callCount).to.be.above(0);
   });
 
-  it('should NOT create layers when the zoom level is outside allowed range', function (done) {
+  it('should NOT draw layers when the zoom level is outside allowed range', function (done) {
 
     map.setZoom(14);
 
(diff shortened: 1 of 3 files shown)


patrickarlt:
@jgravois this looks solid. I made some cleanup changes in https://github.com/Esri/esri-leaflet/pull/697 and all tests pass.

PR #696, opened 2015-12-25
Update bower.json

skarppion101:
Corrected order when using wiredep

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -2,6 +2,9 @@
   "name": "esri-leaflet",
   "version": "v2.0.0-beta.6",
   "main": "dist/esri-leaflet.js",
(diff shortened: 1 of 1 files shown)


patrickarlt:
Thanks @skarppion101!

jgravois:
i think we actually need to explicitly require leaflet `1.0.0-beta.2` instead
@skarppion101's SemVer range only lays down `0.7.7`.

```js
$ npm view leaflet
>>
{ name: 'leaflet',
  description: 'JavaScript library for mobile-friendly interactive maps',
  'dist-tags': { latest: '0.7.7', beta: '1.0.0-beta.2' },
  // ...
```

PR #698, opened 2015-12-27
Use https for all examples

patrickarlt:
@jgravois This is my response to https://github.com/Esri/esri-leaflet/pull/692. Redirecting the site to http just won't work. *https everywhere* isn't just a philosophical stance it is a browser plugin https://www.eff.org/https-everywhere%20 that tries to force everything to use https even at the cost of breaking websites.

The only real way to solve this is to use https for everything so this PR removes all samples that I couldn't find a comparable service too. I also did a full audit of the samples and tweaked a few things and removed a few samples here and there.


site/source/pages/examples/dynamic-map-layer-reprojected.hbs
```diff
@@ -1,20 +0,0 @@
...
-<div id="map"></div>
-
-<script>
-  //Initialize the map to center on Western Europe
-  var map = L.map('map').setView([49.39609, 5.76991], 4);
-
-  //Add Gray Basemap
-  L.esri.basemapLayer("Gray").addTo(map);
-  var dynLayer = L.esri.dynamicMapLayer({
-    url: "http://sampleserver6.arcgisonline.com/arcgis/rest/services/SampleWorldCities/MapServer"
```
jgravois: this service is available over [https](https://sampleserver6.arcgisonline.com/arcgis/rest/services/SampleWorldCities/MapServer)

jgravois: i'll fix when i resolve my merge conflict in #701

jgravois:
this works for me!

PR #699, opened 2015-12-28
bump/syncronize npm/bower leaflet version dependencies

jgravois:
`>=0.7.0` doesn't lay down `1.0.0-beta.x` (yet) because `0.7.7` is still tagged as *latest*.  (more info [here](https://github.com/Esri/esri-leaflet/pull/696#issuecomment-167362232)).

got rid of esperanto dependency while i was at it (to resolve an `npm install` nag)


patrickarlt:
Good catch!

PR #707, opened 2015-12-30
bump virtual grid dep to 1.0.2 explicitly

jgravois:
while this isn't *technically* required, it seems useful to me document in the repo that our next release wouldn't behave correctly with leaflet-virtual-grid `1.0.0`

@ngoldman, do you have opinions about this sort of thing?

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,8 +12,8 @@
   ],
   "dependencies": {
     "arcgis-to-geojson-utils": "^1.0.0",
-    "leaflet-virtual-grid": "^1.0.0",
-    "leaflet": "1.0.0-beta.2"
+    "leaflet-virtual-grid": "^1.0.2",
+    "leaflet": "^1.0.0-beta.2"
   },
   "browserify": {
     "transform": [


patrickarlt:
This is covered in https://github.com/Esri/esri-leaflet/pull/708 so lets just merge that.

ngoldman:
@jgravois @patrickarlt probably a good idea to document compatibility issues in the leaflet-virtual-grid readme.

jgravois:
> probably a good idea to document compatibility issues in the leaflet-virtual-grid readme.

i'm all for that.  i just wondered whether there is some conventional wisdom regarding whether or not you should bother bumping patch numbers in package.json for dependencies when your SemVer range already ensures they will be picked up by a new `npm install`

ngoldman:
I'd say if there's a minimum version for compatibility it's a good idea to set it to that to make it clear versions below that won't work.

jgravois:
well said.  i thank you for your help as always ghost of xmas past.

PR #708, opened 2015-12-30
Use tiny-binary-search module

patrickarlt:
Also updates to latest versions of deps and fixes a few edge cases for time aware feature layers.


jgravois:
confirmed tests are passing and injecting built API into the local website doesn't break any samples.

nice work @patrickarlt!

Issue #691, opened 2015-12-10
Toggling a feature layer on/off quickly causes the visible portion of the layer to remain on the map


bsnider:
See this sample: http://jsfiddle.net/nbn1vgd8/4/

1. Quickly click the addLayer/removeLayer button twice.
2. Zoom out.
3. Note that the layer should be totally removed from the map, but some counties still display.
4. Contrast this with this sample using geojson: http://jsfiddle.net/z3nmxrp3/

jgravois:
this issue [isn't reproducible](http://jsfiddle.net/m976m4wa/) in Leaflet 1.0.0-beta.2 / Esri Leaflet 2.0.0-beta.6, but the repro case does result in a nasty uncaught console error.

![screenshot 2015-12-16 12 05 37](https://cloud.githubusercontent.com/assets/3011734/11852673/2493d008-a3ee-11e5-9dc8-246350250148.png)

this is because internally in the API we call [`FeatureManager._map.getZoom()`](https://github.com/Esri/esri-leaflet/blob/master/src/Layers/FeatureLayer/FeatureManager.js#L150-L153) after the layer has already been removed from the map (and its corresponding private property unset).
```js
var zoom = this._map.getZoom(); // this._map is undefined
```
do any brave souls want to venture a guess as to how we could handle this situation more elegantly?

patrickarlt:
@jgravois I'll take a look at this.

patrickarlt:
@jgravois looks like this is a pre 2.0 issue. Maybe it is related to https://github.com/Esri/esri-leaflet/pull/632?

jgravois:
based on changes introduced in #695, the problem i noted in my first reply within this thread is now completely resolved in `master`.

i'll leave this issue open in case we decide to backport a fix to `1.0.x`

patrickarlt:
@jgravois any chance you could backport to `1.0.x`? It is a verified bug and we did promise to fix bugs.

jgravois:
the fix is [live](http://jsfiddle.net/nbn1vgd8/9/) in `1.0.2`.  let me know if you have any questions @bsnider.

PR #713, opened 2015-12-31
make sure features stop drawing when layer is removed

jgravois:
this is a fairly cheesy solution to #691, but it works fine for me.  it would be a mess trying to backport all the new and improved grid handling logic to `1.x`, 

happy to rework if you have a better idea, if not, i'll bump the internal version numbers to `1.0.2` and push the patch to github/npm.


patrickarlt:
@jgravois `1.0.2` can get released once I merge this.

PR #715, opened 2015-12-31
Update switching-basemaps.hbs

henrygranados:
I thought it would be better to create a function that handles the changing basemap operation instead of using the addEventListener method. It makes it look easier to understand for other developers.
Sincerely;

Henry Granados

Thanks!


jgravois:
sounds good @henrygranados. thx!

PR #723, opened 2016-01-08
get npm install/npm run build working on windows

### LINKED ISSUE #722: 'npm run build' fails on windows
mentioned [here](https://github.com/Esri/esri-leaflet/issues/712#issuecomment-169508877) and confirmed on my own machine.

jgravois:
closes #722
even though we use a [`^`](https://github.com/Esri/esri-leaflet/blob/master/package.json#L63), to denote a SemVer range for rollup (that usually allows bumping the minor), we were still effectively locking the version in place at `0.11.4` because the version number is `< 1.x` (more info [here](https://github.com/npm/npm/issues/6768#issuecomment-64726706)).

i couldn't get the latest (`0.24.0`) working quickly, so this PR just bumps us up to  to `0.19.0`.  i confimed that this has the desired effect on both mac and windows.

### DIFF
--- a/src/Layers/FeatureLayer/FeatureManager.js
+++ b/src/Layers/FeatureLayer/FeatureManager.js
@@ -385,14 +385,15 @@ export var FeatureManager = VirtualGrid.extend({
   _visibleZoom: function () {
     // check to see whether the current zoom level of the map is within the optional limit defined for the FeatureLayer
     if (!this._map) {
-      return false
+      return false;
     }
     var zoom = this._map.getZoom();
-    if (zoom > this.options.maxZoom || zoom < this.options.minZoom) { return false }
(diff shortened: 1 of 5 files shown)


src/Layers/FeatureLayer/FeatureManager.js
```diff
@@ -407,7 +408,7 @@ export var FeatureManager = VirtualGrid.extend({
         var coords = this._activeCells[i].coords;
         var key = this._cacheKey(coords);
         if (this._cache[key]) {
-          this.addLayers(self._cache[key]);
+          this.addLayers(this._cache[key]);
```
jgravois: caught a stray reference to `self` from #695

Issue #728, opened 2016-01-14
FeatureLayerService timeout option


nathanhilbert:
I'm hitting some issues with 2.0.0-beta.7 of featurelayer requests taking up to 5 seconds for certain areas, then returning a 500 from ArcGIS Online.  This is definitely a problem with ArcGIS Online, but I was thinking a workaround for similar slow featurelayerservices would be useful.  

There are two issues that I have seen:

1. I have a featureservice that has a "where" set.  The user clicks on a button, and it updates the "where" to a new query.  If the user clicks on the different buttons quickly, early responses will pend while new calls are being generated and completed.  When that pending call finishes, it triggers the "load" event and draws the geometry of that pended call.  

2. To avoid users from clicking on buttons while loads are happening, I put a flag to ignore clicks while loading.  However, this still hits the issue that if a feature request is pending for five seconds, the user is sitting there waiting.

My first thought was to put a timeout on each "requeststart" event.  If it takes more than 2 seconds, then just abort it.  I can't quite find how to access the httprequest object to call abort.  Maybe someone can point me in the right direction.  Maybe it would be useful to have an option that does this internally and fires necessary events.  Any thoughts would helpful.  Thanks.

abbreviated code:
```
    featureLayer= L.esri.featureLayer({
              url: 'myAGOfeatureserver',
              where: "1=1",
              style: serviceStyle,
              simplifyFactor: 0.1,
              where: "Name='poly001'",
              cacheLayers: false
              });



    featureLayer.on("loading", function(ev){
      currentlyloading = true;
      $("#loading").show();
    });


    featureLayer.on("load", function(ev){
      currentlyloading = false;
      $("#loading").hide();
    });

    featureLayer.service.on("requeststart", function(ev){
      console.log(ev);
      console.log(ev['target']);
     //where's abort?
    });


    featureLayer.addTo(map);

```

jgravois:
i'm hopeful that when we sort out #719 you'll see a performance improvement, but either way, it sounds like a good idea to me t

> Maybe it would be useful to have an option that does this internally and fires necessary events.

what do you mean by 'option'?  my gut reaction to this is that we should ensure all pending requests are aborted when [`featureLayer.setWhere()`](https://github.com/Esri/esri-leaflet/blob/master/src/Layers/FeatureLayer/FeatureManager.js#L188) is called automatically (for everyone all the time).

nathanhilbert:
Thanks for the response.  
[...]

Issue #719, opened 2016-01-04
L.esri.featureLayer is firing duplicate simultaneous queries


jgravois:
#### STR:

1. launch [this sample](http://esri.github.io/esri-leaflet/examples/simple-feature-layer.html)
2. click the '+' button to zoom in once (6 query requests will be issued)
3. click the '+' button to zoom in again

12 requests are issued, 6 of which have identical query geometry.

![screenshot 2016-01-04 12 31 43](https://cloud.githubusercontent.com/assets/3011734/12099728/29c00de4-b2df-11e5-81a4-01f3a9ee88b8.png)

reordered to show duplication (ie: 1 and 7 are identical, 2 and 8, etc.)
```js
"xmin":-122.6953125,   "ymin":45.521743896993634,"xmax":-122.6513671875,"ymax":45.55252525134013
"xmin":-122.6953125,   "ymin":45.521743896993634,"xmax":-122.6513671875,"ymax":45.55252525134013

"xmin":-122.7392578125,"ymin":45.521743896993634,"xmax":-122.6953125,   "ymax":45.55252525134013
"xmin":-122.7392578125,"ymin":45.521743896993634,"xmax":-122.6953125,   "ymax":45.55252525134013

"xmin":-122.6513671875,"ymin":45.521743896993634,"xmax":-122.607421875, "ymax":45.55252525134013
"xmin":-122.6513671875,"ymin":45.521743896993634,"xmax":-122.607421875, "ymax":45.55252525134013


"xmin":-122.6953125,   "ymin":45.49094569262732, "xmax":-122.6513671875,"ymax":45.521743896993634
"xmin":-122.6953125,   "ymin":45.49094569262732, "xmax":-122.6513671875,"ymax":45.521743896993634

"xmin":-122.7392578125,"ymin":45.49094569262732, "xmax":-122.6953125,   "ymax":45.521743896993634
"xmin":-122.7392578125,"ymin":45.49094569262732, "xmax":-122.6953125,   "ymax":45.521743896993634

"xmin":-122.6513671875,"ymin":45.49094569262732, "xmax":-122.607421875, "ymax":45.521743896993634
"xmin":-122.6513671875,"ymin":45.49094569262732, "xmax":-122.607421875, "ymax":45.521743896993634
```

more problematically, none of the requests result in a `304` status indicating browser cache was utilized.  they all `200`.

patrickarlt:
[...]

Issue #706, opened 2015-12-29
disableClientCache


bbinckes:
This was an issue that Jeff G. added 12/22/2015 "How do you disable client caching in L.esri.Layers.DynamicMapLayer?"

The actual issue was related to the QueryTask on a FeatureLayer caching images.

jgravois:
my reply to Jeff G. can be found [here](https://geonet.esri.com/thread/170100)

QueryTasks/FeatureLayers don't return images, but rather JSON features so it'd be really helpful if you are able to provide additional clarification and/or a code snippet to explain in more detail.

bbinckes:
Continued....  Sorry I sent this prematurely.
Background:  The SDE data source that is backing this service is updated every 2 minutes so we need a mechanism to disable cache on the Query. 

I have updated esri leaflet and tested out the changes below which provides the behavior we are looking for but not sure if this is the correct way to approach this.

Added options to the instantiation of a FeatureLayer
MapModel.js
  conditions: L.esri.featureLayer({
    url: dts.config.map.layers.conditions,
    position: 'back',
    disableClientCache: true
  })


Added ts parameter EsriLeaflet.Tasks.Query
Set it to zero, if disableClientCache = false this value will never be updated.

  params: {
    returnGeometry: true,
    where: '1=1',
    outSr: 4326,
    outFields: '*',
    ts: 0
  },


Update timestamp parameter _setGeometry 
Common function that is called in Query Task.  Update parameter if TRUE.
  _setGeometry: function(geometry) {

    if (this.options.disableClientCache) { this.params.ts = Date.now();

Thanks,

jgravois:
that sounds like what i was suggesting.

if you'd like to help us get the functionality into the API for others to use, i'd be happy to review a pull request.  extreme bonus points if you can include relevant documentation updates and some corresponding test coverage as well.

jgravois:
i'm always happy to review a PR which introduces the functionality i sketched out in my reply above if anyone ever wants to take a stab at it.

PR #730, opened 2016-01-15
add timeout option and keep correct count of active requests

nathanhilbert:
See more info at #728.

This adds a timeout option when FeatureLayer is instantiated or the possibility to setTimeout(milliseconds) on the service itself.  In addition, FeatureManager was not keeping an accurate count of activerequests, since any requests that error out would not be post-processed.  Tests were added for both cases.

The test 'keep an accurate count of active requests even when they error' in FeatureManagerSpec could also be useful for #719.

### DIFF
--- a/spec/Layers/FeatureLayer/FeatureManagerSpec.js
+++ b/spec/Layers/FeatureLayer/FeatureManagerSpec.js
@@ -870,4 +870,51 @@ describe('L.esri.FeatureManager', function () {
 
   });
 
+  it('should keep an accurate count of active requests even when they error', function(){
+    var triggered = false;
+
+    layer.on("load", function(ev){
+      triggered = true;
+    });
+
+    var southWest = L.latLng(29.53522956294847, -98.4375),
+        northEast = L.latLng(30.14512718337613, -97.734375),
+        bounds = L.latLngBounds(southWest, northEast);
+    var point = L.point(200, 300);
+
+    server.respondWith('GET', new RegExp(/.*/), JSON.stringify({
+      fields: fields
+    }));
+
+    layer._requestFeatures(bounds, point, function(){return;});
+    
+    server.respondWith('GET', new RegExp(/.*/), JSON.stringify({
+      fields: fields
+    }));
+
+    layer._requestFeatures(bounds, point, function(){return;});
+
+    expect(layer._activeRequests).to.equal(2);
+    
+
+    server.respondWith("GET", new RegExp(/.*/),
+                [500, { "Content-Type": "application/json" },
+                 '']);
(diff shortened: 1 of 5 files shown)


spec/Layers/FeatureLayer/FeatureManagerSpec.js
```diff
@@ -870,4 +870,51 @@ describe('L.esri.FeatureManager', function () {
...
+                 '']);
+    layer._requestFeatures(bounds, point, function(){return;})
+
+    server.respond();
+    server.respond();
+    server.respond();
+
+    expect(layer._activeRequests).to.equal(0);
+    expect(triggered).to.be.true;
+
```
patrickarlt: Could this extra whitespace be cleaned up?

src/Services/Service.js
```diff
@@ -7,7 +7,8 @@ export var Service = L.Evented.extend({
 
   options: {
     proxy: false,
-    useCors: cors
+    useCors: cors,
+    timeout: 5000
```
patrickarlt: Lets keep the timeout option at the `XMLHTTPRequest` default of `0`. Otherwise this feels like a breaking change.

src/Services/Service.js
```diff
@@ -41,6 +42,10 @@ export var Service = L.Evented.extend({
     return this;
   },
 
+  setTimeout: function (timeout) {
```
patrickarlt: Could you implement a companion `getTimeout` route? It is nice to have parity with the getters and setters.

patrickarlt:
Thanks @nathanhilbert for a fantastic PR. I've made a few small comments.

Overall I'm not sure I like the approach of using `context` as an additional options object but I'm not sure there is any other way to do it without a major refactor of the method signatures. This might get problematic when trying to use `context` to pass a custom context to the `callback` and also expecting `timeout` to work. For example:

```
var service = new L.esri.Service({
  timeout: 5000
});

var someUserObject = {
  // ...
}

// since someUserObject is NOT the service with the timeout option this request will not timeout.
service.get(url, params, callback, someUserObject).
```

This wont break any apps so it isn't a breaking change. It just means that `timeout` and `context` cannot be used together. What i like about this approach is that it also gives us a way to implement https://github.com/Esri/esri-leaflet/issues/706 and add cache busting to services in addition to timeouts.

If you can take care of the specific issues I mentioned I'll be more then happy to merge this.

nathanhilbert:
Thanks for the comments @patrickarlt . Suggested changes were added and tests updated.

patrickarlt:
Thanks @nathanhilbert!

PR #740, opened 2016-02-03
RasterLayer.getEvents deleted

SVoyt:
Because in ‘onAdd’ function ‘update’ function already binded to
‘moveend’ event. If ‘getEvents’ exists, ‘update’ is calling two times


jgravois:
i confirmed that `L.esri.dynamicMapLayer` was previously executing its `_update()` method twice.
i couldn't say for sure, but i have a hunch this is a leftover from something we needed in leaflet `0.7.x`.

thanks for the fix!

PR #751, opened 2016-02-26
Fix readme server instructions

hamhands:
### What changed
* Fix server instructions (`npm run start` rather than `npm run serve`) 
* *literally* nothing else

Pedantic PR is pedantic.

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,7 +114,7 @@ Feel free to add your own project to this list!
 1. [Fork and clone Esri Leaflet](https://help.github.com/articles/fork-a-repo)
 2. `cd` into the `esri-leaflet` folder
 3. Install the dependencies with `npm install`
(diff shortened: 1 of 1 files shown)


jgravois:
thanks for this @hamhands! I forgot to update the README when I moved the Esri Leaflet website into http://github.com/Esri/esri-leaflet-doc

i'll see your pedantry (pedantism?) and raise you... 

`npm start` is one of a small handful of 'keyword' npm script commands that doesn't need to be preceded by `run`, so in the interest of saving folks a few keystrokes, (and humoring my impulse to nitpick) would you be game to amend the change slightly in one more commit?

https://docs.npmjs.com/cli/start

hamhands:
Slightly amended @jgravois!

jgravois:
gracias!

Issue #733, opened 2016-01-17
Refactor build process


patrickarlt:
I increasingly don't like shipping a [custom build script](https://github.com/Esri/esri-leaflet/blob/master/scripts/build.js) with every Esri Leaflet repo. It would be able to use https://github.com/Esri/esri-leaflet-bundler to bundle leaflet and all its plugins. Since everything is Rollup-based I don't think this will be too hard.

patrickarlt:
This should also include:

* Upgrade to `rollup@latest`
* Utilize rollup plugins as much as possible
* Add a `--debug` build with unminified code and sourcemaps to resolve https://github.com/Esri/esri-leaflet/issues/712

rowanwins:
Hey @patrickarlt  let me know if you need someone to test on a windows machine

Issue #734, opened 2016-01-19
Refactor module bundling and packaging process


patrickarlt:
While the experiments in shipping ES 6 code for various system has been a pretty big success it is proving too difficult to maintain and given recent changes to babel (and all its implementations) I have been unable to test it shipping ES 6 modules still works with babelify, ect...

The approaches explored in this repo do work, but it is ultimately much simpler to just ship the following:

* Compiled ES 5 UMD module, for CDN, browserify, webpack, AMD. This lives in `main` in the `package.json`
* Compiled ES 6 module with `import/export` for JSPM and Rollup. We will keep the `jspm` and `jsnext:main` fields in `package.json` for compatibility with those systems.

Background info https://github.com/Esri/arcgis-to-geojson-utils/issues/3#issuecomment-167886814

jgravois:
this was closed by #754

PR #754, opened 2016-02-29
Build system and module packaging revamp

patrickarlt:
Fixes https://github.com/Esri/esri-leaflet/issues/733 and https://github.com/Esri/esri-leaflet/issues/734


jgravois:
just pulled all this down and had no trouble running locally.

i'm not sure i understand the benefit of creating both an `esri-leaflet.js` and `esri-leaflet-debug.js` build since i'm able to set breakpoints and debug sourcemaps for the first (even in chrome :100:)

PR #770, opened 2016-04-19
make sure XMLHttpRequest.open is called first

jgravois:
this is a fix to workaround an ie10/11 bug that a customer pointed out in [GIS.SE](http://gis.stackexchange.com/questions/188964/invalidstateerror-in-ie-for-esri-leaflet) that causes the `L.esri.featureLayer` in our [own sample](http://esri.github.io/esri-leaflet/) to fail to draw entirely in those versions of the browser.

the change ensures that we call methods on `XMLHttpRequest` in the following sequence:
1. `open()`
2. `whatever()`
3. `send()`

more or less lifted from google/tracing-framework#407

cc/ @patrickarlt

### DIFF
--- a/src/Request.js
+++ b/src/Request.js
@@ -118,6 +118,16 @@ function xmlHttpGet (url, params, callback, context) {
 export function request (url, params, callback, context) {
   var paramString = serialize(params);
   var httpRequest = createRequest(callback, context);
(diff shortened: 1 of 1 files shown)


patrickarlt:
I hate IE.

PR #773, opened 2016-04-21
get native geojson from 10.4

### LINKED ISSUE #760: get native GeoJSON from ArcGIS Server 10.4
right now we interrogate featureLayer URLs to determine if they are on ArcGIS Online (and can emit GeoJSON natively).

we'll need a metadata request to the service in order to identify 10.4 machines with the same capability.

jgravois:
resolves #760.

i have no idea how, but i've managed to break a bunch of `L.esri.imageMapLayer` tests with this change.  

@tomwayson, do you have time to sit down w/ me and take a look sometime soon?

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -12,11 +12,18 @@ module.exports = function (config) {
     frameworks: ['mocha', 'chai-sinon'],
 
     // list of files / patterns to load in the browser
+    // not sure why tests are failing when files are loaded in bulk
+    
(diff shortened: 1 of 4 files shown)


jgravois:
cc/ @tomwayson

i don't know *why*, but declaring an explicit order when loading tests in `karma.conf.js` is the only thing i needed to do to 'fix' the problem.

PR #784, opened 2016-05-25
default to loading a blank png when a tile request fails

### LINKED ISSUE #759: Broken thumbnails with tile layers
I think I have found a bug:
http://esri-es.github.io/JavascriptAPI/problems/morethan1000entities/tiled_feature_leaflet.html

Cheers,
Raul

jgravois:
resolves #759

it seems sensible to me that we just load a transparent .png by default when individual tile requests return a 404 error and easiest to just grab it directly from the Esri knowledge base article i tracked down.

Article ID:	36939
http://support.esri.com/en/knowledgebase/techarticles/detail/36939

### DIFF
--- a/src/Layers/TiledMapLayer.js
+++ b/src/Layers/TiledMapLayer.js
@@ -4,7 +4,8 @@ import mapService from '../Services/MapService';
 
 export var TiledMapLayer = L.TileLayer.extend({
   options: {
-    zoomOffsetAllowance: 0.1
+    zoomOffsetAllowance: 0.1,
+    errorTileUrl: 'http://downloads2.esri.com/support/TechArticles/blank256.png'
   },
 
   statics: {


blq:
I notice that this blank tile png is hosted on a plain http url. 
Would it be possible for you to host it on https instead? A site that uses https for maps will otherwise block it and get lots of mixed content warnings!

jgravois:
> Would it be possible for you to host it on https instead?

thats a good idea. the current resource doesn't support encrypted requests so i'll have to think a bit about where else to store the `.png`.

blq:
OK. or use an inline base64 image? 

`data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAQAAAAEABAMAAACuXLVVAAAAA1BMVEUzNDVszlHHAAAAAXRSTlMAQObYZgAAAAlwSFlzAAAAAAAAAAAB6mUWpAAAADZJREFUeJztwQEBAAAAgiD/r25IQAEAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA7waBAAABw08RwAAAAABJRU5ErkJggg==`

(the above is exact blank256.png encoded, but I haven't tested live)

patrickarlt:
@jgravois what @blq would suffice pretty nicely without having to host an additional resource.

PR #790, opened 2016-06-24
fire appropriate feature events when entire featureLayer is toggled

### LINKED ISSUE #788: L.esri.FeatureLayer addfeature and removefeature events not fired at all
* Browser and version: Chromium 51.0.2704.84
* Version of Leaflet (`L.version`): 0.7.7
* Version of esri Leaflet (`L.esri.VERSION`): 1.0.3
* I'm loading/bundling the library using: [CDN](https://cdn.jsdelivr.net/leaflet.esri/1.0.3/esri-leaflet.js)

Steps to reproduce the error:
1. Run this [jsbin](http://jsbin.com/mazagipage/1/edit?html,output).
2. In layer control, disable Neighborhoods.

What happens is that labels are kept on map even after the parent layer was disabled.

I was expecting that `addfeature` and `removefeature` could work to add and remove labels when enabling/disabling the parent layer. These events are never fired tough, as even the `console.log` messages aren't throw. Is there a bug with these events?

jgravois:
resolves #788

previously the `addfeature` and `removefeature` events were specific to situations in which an individual feature was appended or dropped from the collection associated with a layer that was currently being drawn (typically because of panning and zooming).

this change extends the logic to fire `removefeature` events for individual features when the entire layer is removed from the map.
for consistency, i also ensured that `addfeature` fires both when a featureLayer is added to the map initially, and again after it is toggled back on.

cc/ @denydias

### DIFF
--- a/spec/Layers/FeatureLayer/FeatureLayerSpec.js
+++ b/spec/Layers/FeatureLayer/FeatureLayerSpec.js
@@ -18,17 +18,17 @@ describe('L.esri.FeatureLayer', function () {
   var layer;
   var map = createMap();
   var features = [{
-        type: 'Feature',
-        id: 1,
-        geometry: {
-          type: 'LineString',
-          coordinates: [[-122, 45], [-121, 40]]
-        },
-        properties: {
-          time: new Date('January 1 2014').valueOf(),
-          type: 'good'
-        }
-      },{
+    type: 'Feature',
+    id: 1,
+    geometry: {
+      type: 'LineString',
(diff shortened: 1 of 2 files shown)


mepdr:
@jgravois with this PR calling setWhere results in firing the addfeature event twice.
see http://jsbin.com/kiripelipa/edit?html,output

jgravois:
thanks for the repro case. please open a new issue for this to help us track the bug report.

PR #795, opened 2016-07-01
Fix esri logo removal check condition

sinaru:
Why `!map._hasEsriLogo`? It causes to return false when there is esri logo in the map. Proposed changed remove the `!` and also return a boolean from the div references. Maybe only `map._hasEsriLogo` is enough.

### DIFF
--- a/dist/esri-leaflet-src.js
+++ b/dist/esri-leaflet-src.js
@@ -1642,7 +1642,7 @@ EsriLeaflet.Tasks.identifyFeatures = function(params){
     },
     onRemove: function(map){
       // check to make sure the logo hasn't already been removed
-      if(!map._hasEsriLogo && this._logo && this._logo._container){
+      if(map._hasEsriLogo && !!(this._logo && this._logo._container)){
         map.removeControl(this._logo);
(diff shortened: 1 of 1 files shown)


jgravois:
good catch! i just pushed up a new [`v1.0.4`](https://github.com/Esri/esri-leaflet/tree/v1.0.4) branch because i plan to backport another [bug](https://github.com/Esri/esri-leaflet/issues/788#issuecomment-228459545) fix soon. 

i can make the same change myself, but i'm also happy to review/merge your own commit if you can make the change in the raw [`src/`](https://github.com/Esri/esri-leaflet/blob/v1.0.4/src/Layers/BasemapLayer.js#L231) instead of the minified build.

sinaru:
Hi mate. You can go ahead and make the change in src/. I just wanted to point out. Encountered this behavior when we used multiple layers and changing the layer. :) However kindly update it asap. Thanks.

jgravois:
[`1.0.4`](https://github.com/Esri/esri-leaflet/blob/master/CHANGELOG.md#104) has been tagged.

PR #797, opened 2016-07-06
less esri in attribution

jgravois:
now that we skip displaying a logo and just add *&copy;Esri* to Leaflet's attribution control, we need to make sure the text isn't duplicated in edge cases where additional credits can't be fetched.

for example:
```js
var map = L.map('map').setView([45, 1080], 5);
L.esri.basemapLayer('Topographic').addTo(map);
```


sinaru:
Ok thanks for the update.

PR #800, opened 2016-07-13
allow users to pass tokens through to basemaps

jgravois:
while it is **not** necessary to provide an API token or other means of authentication to access Esri's hosted basemaps, extremely high volume requests are rate-limited, so we need to provide a mechanism to supply an application token.

once this is merged and a new release is tagged, we'll need to document the new constructor option.


jgravois:
tests are passing locally. merging.

PR #812, opened 2016-07-29
Fixes Issue 807 by adding token to url received from ArcGIS Server

### LINKED ISSUE #807: ImageLayer Token Authentication
Hi All,

For a project I'm working on I needed to authenticate an ImageService using esri-leaflet. I have code to fix this issue if the maintainers would like a pull request (see below).

* Browser and version:

Chrome 51/52, Safari 9.1.2

0.7.7, 1.0.0-rc.1, 1.0.0-rc.2

1.0.3, 2.0.1

Steps to reproduce the error:

1) Create a new html file with the contents of: http://esri.github.io/esri-leaflet/examples/arcgis-server-auth.html

2) Replace L.esri.dynamicMapLayer with L.esri.imageMapLayer

3) Replace URLs and Username/Password with your server and credentials

4) Open file in browser

What happens is:

1)  Browser gets token correctly (200 status)

2) Browser performs GET call to the service's "?exportImage" endpoint with token, bounding box, format, and other parameters

3) Browser receives an object with href, width, height, extent, and scale properties

3) Browser attempts to get the .png image using the href string in the previous response, but it returns a 499 error (Token Required)

I was expecting the service to return the image for leaflet to display.

I'd be happy to make a plunkr if you have an Image Service with public credentials.

I've tried with both CDN and [browserify](http://browserify.org/)

*Fix:*

I discovered that appending the token to the "href" returned by the image service will correctly return the image.

Adding:
if (this.options.token) {
  response.href += ('?token=' + this.options.token);
}

After:
if (error) { return; }

in src/Layers/ImageMapLayer.js seems to work.

Thanks,

Alex

alexpatow:
Fixes #807 by appending a token string with every call from leaflet to the image service.

### DIFF
--- a/src/Layers/ImageMapLayer.js
+++ b/src/Layers/ImageMapLayer.js
@@ -192,6 +192,9 @@ export var ImageMapLayer = RasterLayer.extend({
     if (this.options.f === 'json') {
       this.service.request('exportImage', params, function (error, response) {
         if (error) { return; } // we really can't do anything here but authenticate or requesterror will fire
+        if (this.options.token) { 
+          response.href += ('?token=' + this.options.token); 
+        }
         this._renderImage(response.href, bounds);
       }, this);
     } else {
--- a/src/Layers/ImageMapLayer.js
+++ b/src/Layers/ImageMapLayer.js
@@ -192,8 +192,8 @@ export var ImageMapLayer = RasterLayer.extend({
     if (this.options.f === 'json') {
       this.service.request('exportImage', params, function (error, response) {
         if (error) { return; } // we really can't do anything here but authenticate or requesterror will fire
-        if (this.options.token) { 
-          response.href += ('?token=' + this.options.token); 
+        if (this.options.token) {
+          response.href += ('?token=' + this.options.token);
         }
         this._renderImage(response.href, bounds);
       }, this);


alexpatow:
I'm having issues with npm test. I get all tests passing about 50% of the time. Any thoughts?

Thanks,

Alex

jgravois:
> Any thoughts?

i started seeing a lot more consistent behavior when running tests when i upgraded to Node `v4`.  even though thats the version Travis targets, it can still be a little flaky.  

i restarted the build and everything is green now.

one thing i noticed in my own testing is that tokens don't actually seem to be required when accessing raw images when i test against `10.2`.  what version of ArcGIS Server are you using to host your image services?

https://landscape3.arcgis.com/arcgis/rest/services/USA_Critical_Habitat_Mexican_Spotted_Owl/ImageServer/exportImage?bbox=-13829598.653580371%2C3522218.263380923%2C-10363638.043017337%2C4933551.553638417&size=1417%2C577&format=jpgpng&transparent=true&bboxSR=3857&imageSR=3857&token=JyrbBR8IbjSXJOvQrc8myosWMXnxLoETmjj1IdpCcRdHhDVt551dYN5qPS-9gWyQmz-uObeZNyDwMHQ6gEpynZBe2EcJaxvQ1HUePpWzhsYJuGde6TyQDKZg23oIxoS8XPvht-rqWKkeir-DnVP76g..&f=json

> click on the **href** link included in the response and notice that an image is returned even without a token.

my short-lived token will expire in 24 hours, afterwards, you'll need to generate your own.

instructions can be found in the [Registering your application](https://developers.arcgis.com/authentication/accessing-arcgis-online-services/) section of the following doc page.

alexpatow:
Great! glad the tests are passing, thanks for the heads up.

We're using 10.4.

When we worked with ESRI support on this issue, they mentioned not needing to pass a token when requesting the raw image as well, but apparently that's not how our server is configured, which is what led to this work-around.

PR #828, opened 2016-08-09
Modified featureLayer.unbindPopup to provide features.feature to callback

tyleralves:
Modified featureLayer.unbindPopup method to set this._popup to null rather than false to satisfy test

Used code from esri/esri-leaflet-clustered-feature-layer

### DIFF
--- a/src/Layers/FeatureLayer/FeatureLayer.js
+++ b/src/Layers/FeatureLayer/FeatureLayer.js
@@ -108,6 +108,11 @@ export var FeatureLayer = FeatureManager.extend({
           this.options.onEachFeature(newLayer.feature, newLayer);
         }
 
+        // bind a popup if we have one
+        if (this._popup && newLayer.bindPopup) {
+          newLayer.bindPopup(this._popup(newLayer.feature, newLayer));
+        }
+
         // cache the layer
         this._layers[newLayer.feature.id] = newLayer;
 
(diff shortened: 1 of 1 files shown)


jgravois:
just to reiterate the gist of a discussion that started in Esri/esri-leaflet-doc#61, the purpose of this pull request is to ensure that `featureLayer.bindPopup()` once again passes back a GeoJSON feature, rather than a reference to its parent layer.

i'm still scratching to my head as to why the original code was removed in c2aec4e5c7d7d1398ada745b49480ef6fef639b2.

* our current API reference proclaims, albeit clumsily, that a single GeoJSON feature will be returned by bindPopup()
* the current *actual* behavior is an unintentional deviation from what was seen in esri leaflet `1.0.x` 
* the sole `featureLayer` [sample](http://esri.github.io/esri-leaflet/examples/feature-layer-popups.html) which calls the method incorrectly leads folks to assume that a generic event object is returned (*this one is my fault*).

because of the above my instinct is to call this a *bug fix* instead of labeling it a *breaking change*.  

___
aside:
 i've been wishing a lot lately that Leaflet itself would just tag `1.0.0` and start actually doing ***SemVer*** so the non-hyprocite in me would like to make sure the record states that 'bumping the major' is nothing to be afraid of.

tyleralves:
Correct me if I'm wrong, but Leaflet passes back the reference to the parent layer from featureLayer.bindPopup().

Why then, are we extending their bindPopup method to pass back layer.feature? An alternate option would be to modify (remove?) clusteredFeatureLayer.bindPopup so that they consistently pass back the reference to the parent layer (like Leaflet).

jgravois:
> Why then, are we extending their bindPopup method to pass back layer.feature? 

initially i felt like the best course of action would be to ensure consistency with our historical API, but you raise a fair point that there is also a benefit to not scaffolding unique behavior where none is required.  

if you'd like to create a PR to remove `bindPopup()` from clusteredFeatureLayer instead (and just document the current behavior in esri-leaflet core appropriately), i'm more than happy to go that route.

jgravois:
closed in favor of Esri/esri-leaflet-clustered-feature-layer#8
thanks @tyleralves!

PR #830, opened 2016-08-23
Add in missing token concatenation on DynamicMapLayer image service requests.

jaredbrookswhite:
Missing token concatenation prevented token based image service requests from DynamicMapLayer.


jgravois:
i incorporated a similar patch recently from @alexpatow (see #812).  is there a documented configuration option in ArcGIS Server that i am unaware of that makes it mandatory to pass a token when fetching raw image output?

the sample below shows that it is usually sufficient to pass a token to `/exportMap?`
http://esri.github.io/esri-leaflet/examples/arcgis-server-auth.html

jaredbrookswhite:
According to the client, the only non standard configuration they have on the specific service is standard token based security on the mapservice, running on server 10.3. For the specific app I'm building, the extra line of code to pass the token was necessary to keep the service from returning a code 499 when requesting the image.

jaredbrookswhite:
No problem. Thanks for merging.

Issue #831, opened 2016-08-24
tiledMapLayer should fetch/display service attribution


jgravois:
it wouldn't be hard to handle updating attribution using metadata from tiled map services
✨  *automagically* ✨ 

* if the developer passed `attribution` in the object constructor, don't overwrite it
* if they didn't, we're already making a [`metadata`](https://github.com/Esri/esri-leaflet/blob/master/src/Layers/TiledMapLayer.js#L104-L122) request when the layer is added to the map

```js
// something like this should work...
if (!layer.options.attribution && map.attributionControl) {
  layer.options.attribution = metadata.copyrightText;
  map.attributionControl.addAttribution(layer.getAttribution());
}
```

tyleralves:
I'll take it.

jgravois:
thx @tyleralves

PR #832, opened 2016-08-24
Tfl meta attrib

tyleralves:
Per #831 
Can you take a look at the test I wrote for the auto attribution. It passes if the tiledLayerServer has a _copywriteText_ property, but has not expect whatsoever if not. Is that the proper way to do it?

Cheers,
Tyler

### DIFF
--- a/spec/Layers/TiledMapLayerSpec.js
+++ b/spec/Layers/TiledMapLayerSpec.js
@@ -1,4 +1,17 @@
 describe('L.esri.TiledMapLayer', function () {
+  function createMap(){
+    // create container
+    var container = document.createElement('div');
+
+    // give container a width/height
+    container.setAttribute('style', 'width:500px; height: 500px;');
+
+    // add contianer to body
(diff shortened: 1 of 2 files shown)


spec/Layers/TiledMapLayerSpec.js
```diff
@@ -102,4 +116,31 @@ describe('L.esri.TiledMapLayer', function () {
...
+  it('should display a metadata attribution if one is present and no attribution option was passed', function(){
+    var copywriteText;
+    layer = L.esri.tiledMapLayer({
+      url: url
+    }).addTo(map);
+
+    layer.on('load', function(){
+      layer.metadata(function(err, meta){
+        if(meta && meta.hasOwnProperty(copywriteText)){
+          copywriteText = meta.copywriteText;
```
jgravois: `copywriteText` > `copyrightText`

jgravois:
lgtm!

Issue #700, opened 2015-12-28
should we establish a timeline to drop bower support?


jgravois:
most of bower's original maintainers have moved on, but the project has undergone a rally for funded support lately so proclamations that it is [dead](https://twitter.com/nachocoloma/status/663622545162280960) seem a bit dramatic to me.

i don't use bower, so from my point of view, keeping things in order here to ensure other people can is a [small pain](https://twitter.com/tmcw/status/587687343231934464). not because its a *lot* of work, but because the work needs to be done in
* esri-leaflet-renderers
* esri-leaflet-geocoder
* esri-leaflet-clustered-feature-layer
* esri-leaflet-heatmap-feature-layer
* esri-leaflet-gp
* esri-leaflet-related

you get the idea.

do we want to support installing via bower forever?
should we consider dropping support at `2.0.0` final?
maybe it would be better to add information to the FAQ warning that we'll ditch it at some specific point in the future?
perhaps we could just pass on the maintenance responsibility to those in the community that rely on it?

looking for feedback from anyone/everyone here.

patrickarlt:
As much as a I would love drop bower support i think we need to keep it around. A lot of people requested it initially and that is why it is there. I agree with the reasoning but I think planning is a little premature.

jgravois:
thats good enough for me.  i'll break it up again in 2017 :smile:

PR #834, opened 2016-08-30
add license to bower.json

anotheredward:


jgravois:
thanks for this.  my first instinct is that including a license in the `package.json` should be sufficient since bower ultimately just pulls down a git tag.  

is there a specific reason why i might be wrong?

anotheredward:
Hi! Thanks for your reply.

The reason I've put this PR together is to make scraping licenses easier for automated tooling.
When a project has many bower packages it takes a lot of time to look at each one individually and work out what the licenses are.

If a project has a bower.json, does it also always have a package.json with correct license information?

If so, I have seriously misunderstood bower and I apoligise!

jgravois:
thanks for taking the time to clarify.

> If a project has a `bower.json`, does it also always have a `package.json` with correct license information?

there's no guarantee that a package installed via bower will have a `package.json` (which serves as an indication that it has been published to `npm`) but due to npm's popularity, it'd definitely be smart for you to look for such a file in the root of a dependency's directory in order to glean information about the license.

this is kind of *yet another reason* bower is a source of frustration for library authors and why some [rally against it](https://gofore.com/stop-using-bower/).

as far as esri-leaflet goes, we continue to make our plugins available via bower (#700), but we have been intentional about migrating as much duplicitous information as possible **out** of the `bower.json` file as possible.

hope that helps!

anotheredward:
Thanks for your patience.
You've been very clear on where you and the project stand, I will happily close my PR and tweak my searches :).
I'd love if npm were the only way to get packages, but fully support making things available through bower.

Issue #815, opened 2016-08-01
get tests passing in Node version 6.x


jgravois:
i'm seeing lots and lots of failing tests after upgrading node to version `6.3.1` locally.  the problem can be seen in Travis as well.  i was able to fix a few with simple decimal precision tweaks, but others are more mysterious to me.

branch in progress here - https://github.com/Esri/esri-leaflet/tree/node6

jgravois:
rebased against `master` (to incorporate reorder of individual tests and new linting) and it looks like we've made quite a bit of progress.

now, all 238 tests are executing and the 33 errors are all timeouts.  if i can figure out how to fix one, we should be good to go.

[old log](https://travis-ci.org/Esri/esri-leaflet/builds/158506380) / [new log](https://travis-ci.org/Esri/esri-leaflet/builds/169597183)

jgravois:
it looks like the problem is specific to tests that use a [`RegExp`](https://github.com/Esri/esri-leaflet/blob/20d7d98903e932d3e72af907a4039063f0cc9d0a/spec/Layers/DynamicMapLayerSpec.js#L116-L118) in the mock server response.

jgravois:
`RegExp`s in general aren't problematic.  

the only other thing i've noticed thats peculiar about DynamicMapLayer and ImageMapLayer tests is that their server responses both pass back [base64 encoded images](https://github.com/Esri/esri-leaflet/blob/20d7d98903e932d3e72af907a4039063f0cc9d0a/spec/Layers/ImageMapLayerSpec.js#L54).

jgravois:
i took another look this morning and found that our own failing tests are the result of the fact that L.imageOverlay's `load` event never fires. 

if an [`onerror`](https://github.com/Leaflet/Leaflet/blob/9e5903db405319e462f2d813d0908b2f39493f13/src/layer/ImageOverlay.js#L154-L169) handler is injected, it *is* called, but i wasn't able to glean enough information from the result object to determine what exactly is going wrong.

the same behavior is *not* reproducible when testing `L.imageOverlay` directly using the same base64 encoded image. this has me completely stumped because the [`img.src`](https://github.com/Leaflet/Leaflet/blob/9e5903db405319e462f2d813d0908b2f39493f13/src/layer/ImageOverlay.js#L167) is identical when we instantiate `imageMapLayer`.

```js
it('leaflet dummy test', function (done) {
  var imageOverlay = L.imageOverlay(Image1, bounds);
[...]

PR #841, opened 2016-09-06
make powered by esri omnipresent

### LINKED ISSUE #838: 'Powered by Esri' should always be present in the attribution
currently, we only [tweak](https://github.com/Esri/esri-leaflet/blob/master/src/Layers/BasemapLayer.js#L205) Leaflet's attribution prefix when an Esri basemap is added to the map.

since this attribution should be displayed when **any** content from ArcGIS Online is being drawn, the easiest thing to do is just call `setPrefix()` as soon as this plugin is loaded instead of waiting for a basemap to be added.

jgravois:
resolves #838. i :heart: ES6.

### DIFF
--- a/src/Layers/FeatureLayer/FeatureManager.js
+++ b/src/Layers/FeatureLayer/FeatureManager.js
@@ -1,6 +1,6 @@
 import L from 'leaflet';
 import featureLayerService from '../../Services/FeatureLayerService';
-import { cleanUrl, warn } from '../../Util';
+import { cleanUrl, warn, setEsriAttribution } from '../../Util';
 import VirtualGrid from 'leaflet-virtual-grid';
 import BinarySearchIndex from 'tiny-binary-search';
 
@@ -64,6 +64,9 @@ export var FeatureManager = VirtualGrid.extend({
    */
 
   onAdd: function (map) {
+    // include 'Powered by Esri' in map attribution
+    setEsriAttribution(map);
+
     // check to see whether service is 10.4 or above (and can emit GeoJSON natively)
     this.service.metadata(function (error, metadata) {
       var supportedFormats = metadata.supportedQueryFormats;
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,13 @@
 language: node_js
 sudo: false
+notifications:
+  email: false
 node_js:
   - "4"
 cache:
   directories:
     - node_modules
+before_script:
+  - npm prune
+before_install:
+  - npm i -g npm@^3.0.0
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -13,7 +13,7 @@ module.exports = function (config) {
 
     // list of files / patterns to load in the browser
     // not sure why tests are failing when files are loaded in bulk
-    
+
     files: [
       'node_modules/leaflet/dist/leaflet.css',
       'node_modules/leaflet/dist/leaflet-src.js',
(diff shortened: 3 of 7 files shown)


src/Layers/FeatureLayer/FeatureManager.js
```diff
@@ -64,6 +64,9 @@ export var FeatureManager = VirtualGrid.extend({
    */
 
   onAdd: function (map) {
+    // include 'Powered by Esri' in map attribution
+    Util.setEsriAttribution(map);
```
ta2edchimp: Shouldn't this line read `setEsriAttribution(map)` only, as you're explicitly importing `setEsriAttribution`?

jgravois: yes it should. thx! :wink:

ta2edchimp:
Do you have any idea why the travis build fails?
I fetched your branch, pushed it to my fork and [triggered a build which completed successfully](https://travis-ci.org/ta2edchimp/esri-leaflet/builds/158221498).  
Only reason I can think of would be a corrupted build cache or sth like that (the only related difference I spotted between this repo's `.travis.yml` and [those I'm used to](https://github.com/ta2edchimp/opt-cli/blob/master/.travis.yml#L15-L16), is that I run `npm prune` as `before_script`, as the `node_modules` directory is added to travis' cache).

jgravois:
> Do you have any idea why the travis build fails?

i really wish i did. 

i've tried a couple times lately, and manually blowing away the cache doesn't typically resolve the problem.  to make things even more strange, i was able to dependably *break* the build both locally and on travis by pointing at Node v6.x (see #815) and see green when i reverted to v4.x.  but now i'm seeing consistently reproducible errors (even locally) when targeting v4.5.0.

i'll try injecting `npm prune` and see if it helps.  either way, i appreciate the suggestion!

ta2edchimp:
I hope it helps, or at least somehow points into the right direction.
(I originally stumbled over this PR when I was looking into the node6 branch.)

ta2edchimp:
Another thing that just came into my mind:
We experienced serious issues using npm@2.x (for example, over at [`ghooks`](https://github.com/gtramontina/ghooks/issues/71)), as it does some really strange things during npm's post/preinstall hooks and got used to add
```
before_install:
  - npm i -g npm@^3.0.0
```
to our travis configs. Idk if this is related, but I would give it a try (as the failing builds seem somehow random here, too).

jgravois:
still seeing flaky behavior, but either way i appreciate the help.
makes me even more glad to learn there's a flag to turn off email notifications!

ta2edchimp:
Maybe contacting Travis support can bring some insight. They usually respond fast ...

ta2edchimp:
Could you change karma's log level to something more verbose?
I'm irritated not only about the failing tests (that look somehow random), but also by the number of tests run (in the Travis builds, not even half of the tests are run — reminds me of PhantomJS' memory or more general issues on Linux from some time ago, although those errors should be printed out, given even the current log level).

jgravois:
after pushing a commit that sets the logging level to `verbose`, the build is green again.  the next time i see irregular behavior i'll take your advice and contact Travis Support.

ta2edchimp:
At least, it's green! :tada: 
I'm curious if this makes the other PR work, too.

jgravois:
> At least, it's green! 🎉 

absolutely.  i really appreciate your persistence because we've been seeing stuff in Travis that i don't completely understand for a *long* time. 

> I'm curious if this makes the other PR work, too.

i'll rebase it on top of this branch so we can see 😬

jgravois:
merging.  

much thanks to @ta2edchimp for his suggestions related to getting travis to behave.  i'll observe the behavior a bit and let you know how we get on soon.

Issue #840, opened 2016-09-06
make attribution expandable when an ellipsis is displayed


jgravois:
my humble opinion is that Leaflet's attribution control should never exceed a single line in height and that we should be able to expand the text via a mouseclick if an ellipsis is being used for truncation.

jgravois:
closed via 3f0af3c5331be2f086270c4953afebcc56af792a

PR #842, opened 2016-09-07
insert copyright info into leaflet attribution control for all layers automatically

### LINKED ISSUE #837: make sure esri leaflet fetches/displays service attribution automatically
**all** layers should fetch/display service attribution

- [X] `tiledMapLayer` (done in #831)
- [ ] `dynamicMapLayer`
- [ ] `imageMapLayer`
- [ ] `featureLayer` (which would be inherited by `Cluster.featureLayer` etc.)

jgravois:
closes #837, #840

* interrogate `dynamicMapLayer`, `imageMapLayer`, and `featureLayer` for copyright text in service and add it to the control automatically (unless custom attribution is supplied)
* trim attribution to slightly less than the width of the map (and update when the map is resized) using an ellipsis to denote overflow
* make the attribution expandable (and collapsible) when the user hovers over the control.

~~this includes commits from #841.  i'm planning on a rebase/squash after that one lands.~~

### DIFF
--- a/src/Layers/BasemapLayer.js
+++ b/src/Layers/BasemapLayer.js
@@ -1,7 +1,6 @@
 import L from 'leaflet';
 import { pointerEvents } from '../Support';
 import {
-  calcAttributionWidth,
   setEsriAttribution,
   _getAttributionData,
   _updateMapAttribution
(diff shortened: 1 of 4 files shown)


blq:
It seems the attribution width calculation doesn't correctly take other map controls into account. I use the scale-control for example and it now ends up partially below the attribution text.

jgravois:
@blq thanks for bringing this to my attention.  i was under the *mistaken* impression that other bottom oriented controls would automatically be positioned above the attribution.

i've proposed #849 to resolve the problem. in the meantime, you could inject something like this in your own code.

```js
var attribution = document.getElementsByClassName('esri-truncated-attribution')[0]
// make sure the attribution is trimmed to reserve 150px for other controls
attribution.style.maxWidth = (map.getSize().x - 150) + 'px';
```
you could even execute the code above each time the map is resized if you'd like to retain the dynamic trimming the API provides.

blq:
Ok, thanks, I'll give it a try!

blq:
I tried the new 2.0.4 and I think the attributionWidthOffset margin of 55px is still a bit too low for the scale-control (if in lower left).  I'd say it has to be doubled, i.e more close to your example nr above in this thread! Or use calculations that take controls into account explicitly.

PR #849, opened 2016-09-20
make max-width of attribution control configurable

jgravois:
in the majority of applications it makes sense to leverage as much horizontal space as possible to display attribution, primarily because of small form factor devices, but `55px` is a little safer default because it accommodates a bottom oriented zoom control automatically.

since other possibly bottom aligned controls like L.Control.Scale and L.Controls.Layers don't have a static predictable width our best bet is to just expose a mechanism to let developers indicate just how much space should be reserved.

~~i ***think*** its kosher to just tack on a new `L.Map` control constructor option and introspect for it ourselves later. if so, we'll need to find a new location in our own API reference to document it.~~

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,6 +26,7 @@ coverage
 # Debugging Files
 working
 debug/*.html
+debug/archive/
 !debug/sample.html
 
 # Docs Build
--- a/package.json
+++ b/package.json
@@ -71,7 +71,7 @@
   },
   "scripts": {
     "build": "rollup -c profiles/debug.js & rollup -c profiles/production.js",
-    "lint": "semistandard src/**/*.js | snazzy",
+    "lint": "semistandard \"src/**\" \"src/**/*.js\" | snazzy",
     "prebuild": "mkdirp dist",
     "prepublish": "npm run build",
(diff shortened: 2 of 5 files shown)


jgravois:
@tomwayson reminded me that a generic `L.esri.Options` object would be a more appropriate place to stash a configurable setting than asking folks to start supplying custom constructor options to other API's objects.

the snippet below is an example of how developer's can expect to customize the amount of space reserved for other controls.

```js
// number of pixels to crop attribution width
L.esri.Options.attributionWidthOffset = 200;
var map = L.map('map').setView([37.79, -122.45], 13);
```

*technically*, the proposed implementation could lead to unexpected behavior if a developer tried to pass unique values to different resizable maps as they were created (because the value will be different when `map.on('resize'...` is triggered later.

to be honest, i'm perfectly fine with that.

tomwayson:
A little late on this, but LGTM! Only change I would have suggested is naming it `L.esri.options` w/ lower case "o" just cause it's not an object you can `new` up, but I ain't mad atcha.

jgravois:
agreed.  i'll push up soon.

PR #858, opened 2016-09-28
use L.esri.options (instead of L.esri.Options)

jgravois:
* added tests

### DIFF
--- a/spec/optionsSpec.js
+++ b/spec/optionsSpec.js
@@ -1,0 +1,50 @@
+describe('L.esri.options', function () {
+  function createMap() {
+    // create container
+    var container = document.createElement('div');
+
+    // give container a width/height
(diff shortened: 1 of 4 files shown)


jgravois:
while writing a few hacky tests to inspect the css rules associated with `.esri-truncated-attribution` i have started to suspect that the flakiness in our test suite might be the result of individual tests trampling on one another.

ultimately i needed to loop through and clear a big pileup of stylesheets associated with the document in order to be able to reference them via a consistent index.  i also noticed intermittent local errors in ImageMapLayer tests (which utilize the same test setup and teardown) complaining that an instance of 'Map' was already in use

* i played around limiting the `concurrency` property in karma's configuration file (w/o success)

i'm seeing green consistently locally, and everything passed on Travis, so i'll just keep what i learned in mind for the future.

PR #859, opened 2016-09-28
lint code in test suite

### LINKED ISSUE #855: start linting test directory
semistandard spec/** currently returns ***lots** of errors.

i have a hunch that automatic fixes and `/* eslint-env mocha */` might be all we need.

https://github.com/feross/standard#what-about-mocha-jasmine-qunit-etc

jgravois:
resolves #855

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -26,8 +26,7 @@ module.exports = function (config) {
     ],
 
     // list of files to exclude
-    exclude: [
-    ],
+    exclude: [],
 
     // preprocess matching files before serving them to the browser
     // available preprocessors: https://npmjs.org/browse/keyword/karma-preprocessor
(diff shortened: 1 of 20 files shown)


jgravois:
since there are enough line changes to make anyone's eyes glaze over... this is what was involved:

* declared a few globals in a new `semistandard` property within the package.json
* flagged mocha files individually
* ran `semistandard --fix` to automate the bulk of the corrections
* used eslint ignore comments to suppress complaints about unhandled errors (this is safe because we're testing the API not the callbacks)
* cleaned up actual linting errors manually (there were more than a dozen)

i also rearranged the order of imported test files in karma (again) after noticing that they are usually the source of sporadic errors.  in my own local testing, loading them first resulted in a significant improvement, so i'm hopeful that the same will be true for travis.

since the changes are pretty innocuous and target many different files, i'm going to go ahead and merge this myself now to avoid merge conflicts with other incoming PRs.

PR #871, opened 2016-10-19
Fix #870: layerDefs serialize

### LINKED ISSUE #870: DynamicMapLayer layerDefs not getting serialized properly.
Should only call `JSON.stringify(layerDefs)` if `layerDefs` is not a string. Otherwise it adds literal quotes to the value which is encoded in the resulting url query string.

i.e.
```console
JSON.stringify('VALUE>10')
> '"VALUE>10"'
```

change: 
```js
if (this.options.layerDefs) {
    params.layerDefs = JSON.stringify(this.options.layerDefs);
}
```
to
```js
if (this.options.layerDefs) {
    params.layerDefs = typeof this.options.layerDefs === 'string' ? this.options.layerDefs : JSON.stringify(this.options.layerDefs);
}

```
here:
https://github.com/Esri/esri-leaflet/blob/2cf1b3b375f38a174b9b48544dfbffd2320031e4/src/Layers/DynamicMapLayer.js#L138

whyvez:
fixes #870

### DIFF
--- a/spec/Layers/DynamicMapLayerSpec.js
+++ b/spec/Layers/DynamicMapLayerSpec.js
@@ -263,7 +263,7 @@ describe('L.esri.DynamicMapLayer', function () {
     server.respond();
   });
 
-  it('should get and set layer definitions', function (done) {
+  it('should get and set a JSON layer definitions', function (done) {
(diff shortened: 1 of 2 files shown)


jgravois:
thanks for this @whyvez!  

in the future, you can skip logging an issue and just submit a PR that describes the problem *and* proposed solution if it'd be easier.

> ... layer is instantiated within `beforeEach`. Should I simply add a separate describe or it block and define my own layer?

exactly.  just follow the pattern [here](https://github.com/Esri/esri-leaflet/blob/0c3633b69171460926f403504e809e0486200222/spec/Layers/DynamicMapLayerSpec.js#L79-L83) and holler if you run into trouble.

whyvez:
@jgravois Looks like an issue with the url encoding. Wrote test but its failing. Will tackle later tonight or in the AM.

whyvez:
@jgravois I updated the existing `layerDefs` to deal with a JSON arg and added a similar test for the string arg. Hope this works.

jgravois:
thanks again for this fix @whyvez!

PR #884, opened 2016-11-10
make sure withCredentials request header is set to true in CORS requests

### LINKED ISSUE #881: Possible to set withCredentials to true?
We have a client that has ArcGIS 10.2.2 for Server set up within a secured Active Directory.  We are attempting to add layers from this server to a Mapbox map, but are getting errors.  

Here is the JavaScript code to declare the layer:
```js
var layer = L.esri.dynamicMapLayer({
  url: url,
  opacity: 1,
  layers: layers,
  token: token
});
```

The customer used Fiddler to compare the XHR requests that are being sent from our site against a 3rd party software that is able to access their map data and they stated that the difference was withCredentials was set to true coming from the 3rd party software.

Is it possible to set this by passing another option?

jgravois:
*should* resolve #881

### DIFF
--- a/spec/RequestSpec.js
+++ b/spec/RequestSpec.js
@@ -93,6 +93,17 @@ describe('L.esri request helpers', function () {
     requests[0].respond(200, { 'Content-Type': 'text/plain; charset=utf-8' }, JSON.stringify(sampleResponse));
   });
 
+  it('should make a request with a token', function (done) {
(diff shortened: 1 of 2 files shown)


jgravois:
tested a bit more today and realized that the benefit of setting `withCredentials` isn't to do with hardcoded tokens, but rather to allow cookies like the ones integrated windows authentication generate automatically in the browser are passed through.  i also noticed that this is peculiar to CORS, so its not appropriate to set the header when the browser doesn't support it.

PR #885, opened 2016-11-11
dont convert extent objects with 'NaN' coordinate values to LatLngBounds

### LINKED ISSUE #879: Query bounds Mapserver
Quering a Mapserver layer service for bounds like:
```
query.where("1=2").bounds(function(error, latLngBounds, response){
    map.fitBounds(latLngBounds);
});
```
I got
```
Uncaught Error: Invalid LatLng object: (NaN, NaN)(…)L.LatLng 
@ LatLng.js:24L.latLng 
@ LatLng.js:128T 
@ Util.js:32(anonymous function) 
@ Query.js:139(anonymous function) 
@ Service.js:96s.onreadystatechange 
@ Request.js:75
```
The server response is: `{"extent":{"xmin":"NaN","ymin":"NaN","xmax":"NaN","ymax":"NaN"}}`
The same against a featureServer, (i,e, https://services.arcgis.com/rOo16HdIMeOBI4Mb/arcgis/rest/services/stops/FeatureServer/0/query?returnGeometry=true&where=1=2&outSr=4326&outFields=*&returnExtentOnly=true&f=json) returns 
`{"extent":null}`

Running arcgis 10.4
Live example http://jsbin.com/tatadejiya/1/edit?html,console,output

jgravois:
resolves #879

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,3 +8,6 @@ before_script:
   - npm prune
 before_install:
   - npm i -g npm@^3.0.0
+cache:
+  directories: # Cache dependencies
+    - node_modules
--- a/spec/Tasks/QuerySpec.js
+++ b/spec/Tasks/QuerySpec.js
@@ -193,6 +193,15 @@ describe('L.esri.Query', function () {
     }
   };
 
+  var sampleNaNExtentResponse = {
+    'extent': {
+      'xmin': 'NaN',
+      'ymin': 'NaN',
+      'xmax': 'NaN',
+      'ymax': 'NaN'
+    }
+  };
+
   var sampleCountResponse = {
     'count': 1
   };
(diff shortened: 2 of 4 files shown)


patrickarlt:
@jgravois I would actually argue that instead of making `latlngbounds` in the callback be `null`, we should make `error` be something like `Invalid Bounds`.

This way I can ALWAYS handle `latlngbounds` as a bounds instead of having to do things like:

```js
task.bounds(function (error, latlngbounds, raw) {
  if (error && latlngbounds) {
    // now I have to check for bounds and an error :(
  }
});
```

I can do

```js
task.bounds(function (error, latlngbounds, raw) {
  if (error) {
    // request error or invalid bounds....
    return;
  }

  // now I know I have a valid bounds
});
```

jgravois:
this is good feedback and incorporated. 🚢 ing it.

PR #893, opened 2016-11-23
make sure addfeature is only called on readd

### LINKED ISSUE #886: Calling setWhere fires the addfeature event twice.
Calling setWhere results in firing the addfeature event twice. (seems related to https://github.com/Esri/esri-leaflet/pull/790)
Example: http://jsbin.com/kiripelipa/edit?html,output

* Browser and version:

Chrome 54

1.0.1

2.0.4

jgravois:
resolves #886

### DIFF
--- a/spec/Layers/FeatureLayer/FeatureLayerSpec.js
+++ b/spec/Layers/FeatureLayer/FeatureLayerSpec.js
@@ -143,10 +143,30 @@ describe('L.esri.FeatureLayer', function () {
   });
 
   it('should add features back to a map', function () {
-    layer.removeLayers([1]);
+    layer = L.esri.featureLayer({
+      url: 'http://gis.example.com/mock/arcgis/rest/services/MockService/MockFeatureServer/0',
+      timeField: 'time',
+      pointToLayer: function (feature, latlng) {
+        return L.circleMarker(latlng, {
+          color: 'green'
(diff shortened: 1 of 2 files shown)


patrickarlt:
@jgravois I feel like we should prevent the problem (why is the event firing twice) rather then apply a solution (force feature to only be added once). I can try to find some time to debug why this happens.

jgravois:
agreed.

I took a stab at that initially but i couldn't quite wrap my brain around the internal logic and sequence of events when a fresh filter is applied in a *real world* scenario.

when I updated an existing test to start counting the number of events fired I expected it to trap the bug the user reported, but the issue isn't reproducible when raw features are fed in manually via `createLayers()` at all.

patrickarlt:
@jgravois Looks like @goldenlimit might have figured this out... https://github.com/Esri/esri-leaflet/issues/886#issuecomment-265821830

jgravois:
@goldenlimit was right.

the best course of action is to stop firing `addfeature` when layers are created entirely. this allows us to support the use case described in #788 (*and* avoid the bug reported in #886).

since i never updated the API reference documentation to mention that `addfeature` *was* temporarily triggered when loading features for the first time, arguably what i'm now proposing shouldn't even be considered a breaking change.

PR #904, opened 2016-12-20
honor leaflet maxNativeZoom

### LINKED ISSUE #743: Tiles disappear when zooming beyond 'maxNativeZoom'
After the 'maxNativeZoom' level set on a tile loaded with **L.esri.tiledMapLayer** is reached, the tiles disappear instead of being loaded from the maxNativeZoom level and auto-scaled.

If the tile is loaded using the Leaflet **L.tileLayer** 'maxNativeZoom' works correctly.

I've prepared a [demo](https://jsfiddle.net/6b7tq6ea/2/) to see the issue.

Two tiles have been added using the ESRI Leaflet addon and again using just Leaflet. After zooming in beyond the maximum zoom level where the tile has data (15) the ESRI tiles disappears whereas the Leaflet tile is scalled properly.

jgravois:
resolves #743

@patrickarlt was right.  i just had to make sure and update the logic down in `createTile()` too.

### DIFF
--- a/src/Layers/TiledMapLayer.js
+++ b/src/Layers/TiledMapLayer.js
@@ -63,7 +63,7 @@ export var TiledMapLayer = L.TileLayer.extend({
   getTileUrl: function (tilePoint) {
     return L.Util.template(this.tileUrl, L.extend({
       s: this._getSubdomain(tilePoint),
-      z: (this._lodMap && this._lodMap[tilePoint.z]) ? this._lodMap[tilePoint.z] : tilePoint.z, // try lod map first, then just defualt to zoom level
+      z: (this._lodMap && this._lodMap[this._getZoomForUrl()]) ? this._lodMap[this._getZoomForUrl()] : tilePoint.z, // try lod map first, then just default to zoom level
       x: tilePoint.x,
       y: tilePoint.y
     }, this.options));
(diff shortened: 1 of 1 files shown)


chriserickson:
Just glancing at the patch, did you want to default back to point.z with no lodmap, or default back to the zoom level as the comment states?

eg 
` (this._lodMap && this._lodMap[this._getZoomForUrl()]) ? this._lodMap[this._getZoomForUrl()] : this._getZoomForUrl()`

instead of
` (this._lodMap && this._lodMap[this._getZoomForUrl()]) ? this._lodMap[this._getZoomForUrl()] : tilePoint.z`

jgravois:
good catch! with no lodmap, we should default back to the zoom level.

chriserickson:
I manually hacked this in to our esri-leaflet-src, and it appears to work.

jgravois:
travis will show green here as soon as #906 is merged and the build is restarted.

ahalota:
Should this be working now? I am linking to the file shown on the main page for cdn (https://unpkg.com/esri-leaflet@2.1.1/dist/esri-leaflet-debug.js) but still not getting native zoom respected. Which version of leaflet did you have this working with?

jgravois:
hi @ahalota! the fix was included in [`v2.0.7`](https://github.com/Esri/esri-leaflet/releases/tag/v2.0.7) and continues to resolve the original repro case.

1. launch https://jsfiddle.net/jagravois/z7mjvgvm/
2. whether you toggle the esri layer or the native leaflet layer, both start to resample at zoom 16.

if you have more information or can provide a demo of your own i'd be happy to take another look.

ahalota:
Never mind. It was an odd issue in my ArcServer that was not allowing me to access the higher resolution images and turning up transparent tiles instead...

Issue #908, opened 2017-01-03
setOpacity on RasterLayers


Saulzi:
* Browser and version:

ie: NA

* Version of Leaflet (`L.version`):

1.02

* Version of esri Leaflet (`L.esri.VERSION`):

2.0.6

Steps to reproduce the error:

1. Create a layer which inherits RasterLayer
2. Immediatly set its opacity

What happens is A exception is raised because this._currentImage is null

I was expecting everything to be fine and dandy.

fix is to wrap this._currentImage in if statement.

jgravois:
99% sure this was fixed in #903.

Saulzi:
Nope. basically issue is in rasterlayer.js

```js
  setOpacity: function (opacity) {
    this.options.opacity = opacity;
    this._currentImage.setOpacity(opacity);
    return this;
  },
```
should be
```js
  setOpacity: function (opacity) {
    this.options.opacity = opacity;
    if (this._currentImage)
    {
        this._currentImage.setOpacity(opacity);
     }
    return this;
  },
```

I will shortly have a pr with fix and test,

jgravois:
> I will shortly have a pr with fix and test.

that would be fantastic. :cookie: thanks so much!

jgravois:
resolved by 10666308c018888f614974284a91c5f743d56eba

PR #909, opened 2017-01-03
setOpacity on RasterLayers

Saulzi:
fix for
#908

### DIFF
--- a/spec/Layers/RasterLayerSpec.js
+++ b/spec/Layers/RasterLayerSpec.js
@@ -1,0 +1,9 @@
+/* eslint-env mocha */
+/* eslint-disable handle-callback-err */
+describe('L.esri.RasterLayer', function () {
+  it('should not error when calling setOpacity when _currentImage is null', function () {
+    var layer = new L.esri.RasterLayer();
+    layer._currentImage = null;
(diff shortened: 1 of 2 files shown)


jgravois:
thanks for this!  

i noticed that the commit here is actually associated with github.com/[mrsaulzi](https://github.com/mrsaulzi). this is likely because the email address in your git config is different than the one you used when you created github.com/[Saulzi](https://github.com/mrsaulzi).

> New commits aren't using the right name
https://help.github.com/articles/setting-your-username-in-git/#new-commits-arent-using-the-right-name

i'm certainly happy to accept as is, but if you want to tidy things up so that *Saulzi* appears in perpetuity as the :circus_tent: 50th :tada: esri leaflet [contributor](https://github.com/Esri/esri-leaflet/graphs/contributors), please don't be shy.

(i only mention it because i did exactly the same thing a long time ago)

Saulzi:
I have deleted the mrsaulzi account and added / set email address on this account. does it now appear correctly?

jgravois:
> does it now appear correctly?

yup.  now the commit avatar and pull request avatar are the same. thanks again for this fix!

PR #921, opened 2017-01-31
reverse simplification scale and ensure bindPopup simplifies output geometry

### LINKED ISSUE #919: Add contorl for `returnGeometry` to `dynamicMapLayer.bindPopup` to reduce (unnecessary) load
* Browser and version:

Chrome 55, IE11

1.0.3

2.0.7

Steps to reproduce the error:

1. layer published with large and complicated features from ArcGIS server
2. add layer as `DynamicMapLayer` in leaflet
3. add `bindPopup` to layer and return ~~feature's label `featureCollection.features[0].properties.Label `~~ `abc` as popup content
4. click feature

What happens is [long response time caused by `GET`-ing entire GeoJson object from server].

I was expecting [to get feature's properties alone without getting the whole `geometry` set. User should at least be able to choose if they want to get `geometry`. By choosing to load the layer as `dynamicMapLayer`, some (at least me) were avoiding prolonged loading time & complex svg's building time].

( [jsbin](http://jsbin.com/caperoqimo/1/edit?html,js,output) that demonstrates the problem, click on the polygon near City of Houston and you will recreate the freeze)

* Optional: I'm *not* using the [CDN](http://www.jsdelivr.com/projects/leaflet.esri), I'm loading/bundling the library using: [webpack](https://webpack.github.io/) but the jsbin example was set up using CDN.

jgravois:
resolves #919 (partially anyway)

this patch applies a simplification factor of 0.5 to output geometries in dynamicMapLayer.bindPopup using the current scale of the map to ensure that no more vertices than necessary are retrieved from the server.

for the test case provided by the user, this reduced the payload from `213kb/600ms` to `922b/150ms` when zoomed out pretty far on a complex feature.

screenshot 2017-01-31 12 21 28

i think this is a sensible default so that we can ensure in more cases that the map click interaction is snappy.

### DIFF
--- a/spec/Layers/DynamicMapLayerSpec.js
+++ b/spec/Layers/DynamicMapLayerSpec.js
@@ -374,7 +374,7 @@ describe('L.esri.DynamicMapLayer', function () {
   });
 
   it('should bind a popup to the layer', function () {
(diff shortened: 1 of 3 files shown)


jgravois:
in testing this change i noticed that the simplification scale for `Query` increases from 0 to 1 as more generalization is applied to the geometry.

i've added another commit here to ensure that `IdentifyFeatures` is consistent with Query (and its own documentation).

PR #924, opened 2017-02-09
Add check for layer visibility when cellEnter is called.

keithpower:
If you have a layer that is visible at say zoom 16 - 20 and then zoom out to 15 the layer will not be visible (as expected).  If you then pan out of the current virtual grid bounds and then pan back to your original position the layer becomes visible even though you are still at zoom 15.

This is fixed by checking for visibleZoom within cellEnter so features are only added to the map if in a visible zoom.

### DIFF
--- a/src/Layers/FeatureLayer/FeatureLayer.js
+++ b/src/Layers/FeatureLayer/FeatureLayer.js
@@ -153,7 +153,7 @@ export var FeatureLayer = FeatureManager.extend({
   },
 
   cellEnter: function (bounds, coords) {
-    if (!this._zooming && this._map) {
+    if (this._visibleZoom() && !this._zooming && this._map) {
       L.Util.requestAnimFrame(L.Util.bind(function () {
         var cacheKey = this._cacheKey(coords);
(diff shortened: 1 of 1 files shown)


jgravois:
very cool! 

it looks like either the email or username in your local git config doesn't match what you're using for your github account. its not crucial, but if you want to take the time to clean it up the commit in the PR will be associated correctly.

screenshot 2017-02-09 11 18 12

https://help.github.com/articles/setting-your-email-in-git/

keithpower:
Ok, thanks.  I figured I was bound to mess something up.  I think I fixed it.  Let me know if it looks good on your end.

jgravois:
> Let me know if it looks good on your end.

i'm just making sure the avatar in the thread matches the avatar for the commit.  looks great!

as for the code fix, i'm in the middle of something at the moment, but i'll be able to review more fully (and hopefully write a few tests) soon.  thanks for the contribution!

keithpower:
Great, thanks.

jgravois:
thanks again for the contribution!

PR #936, opened 2017-03-11
eachActiveFeature() to return only layers currently in view

### LINKED ISSUE #933: should FeatureLayer.eachFeature() be limited to returning the current snapshot?
or should we leave that method alone and add `eachActiveFeature()`?

more info: 

patrickarlt [Feb 13] 
@jgravois @reames Looks like `eachFeature` is looping over each layer in the `_layer` cache which will be all layers. https://github.com/Esri/esri-leaflet/blob/master/src/Layers/FeatureLayer/FeatureLayer.js#L246-L251
GitHub
Esri/esri-leaflet
esri-leaflet - A lightweight set of tools for working with ArcGIS services in Leaflet.
 

jgravois [Feb 13] 
i see then.  so if the where filter is added later...

jgravois [Feb 13] 
might be worth adding a programmatic option to clear the local cache

patrickarlt [Feb 13] 
@jgravois the other option would be to have an "each active layer" that used `this._currentSnapshot`

patrickarlt [Feb 13] 
`eachFeature` is "do this for every feature the layer knows about ever."

patrickarlt [Feb 13] 
might be easier to make a new method that is "do this for every feature that is currently shown on the map"

jgravois [Feb 13] 
i know its bad to break things, but i bet everyone expects eachFeature to be `_currentSnapshot`

jgravois [Feb 13] 
i wonder if itd be better to map it that way and add a new `eachFeatureEver()`

jgravois:
resolves #933.

not only did `_currentSnapshot` not already isolate features currently in view, i found that it just was just growing in size as duplicate keys were appended throughout the browser session.

@keithpower i'd be curious to see how the approach i'm proposing here compares to your own technique if you can spare the time to take a look.

### DIFF
--- a/src/Layers/FeatureLayer/FeatureLayer.js
+++ b/src/Layers/FeatureLayer/FeatureLayer.js
@@ -243,6 +243,19 @@ export var FeatureLayer = FeatureManager.extend({
    * Utility Methods
    */
 
+  eachActiveFeature: function (fn, context) {
+    // figure out roughly which layers are in view
+    if (this._map) {
+      var activeBounds = this._map.getBounds();
+      for (var i in this._layers) {
+        if (activeBounds.intersects(this._layers[i].getBounds())) {
+          fn.call(context, this._layers[i]);
+        }
+      }
+    }
+    return this;
+  },
+
   eachFeature: function (fn, context) {
     for (var i in this._layers) {
       fn.call(context, this._layers[i]);
--- a/src/Layers/FeatureLayer/FeatureLayer.js
+++ b/src/Layers/FeatureLayer/FeatureLayer.js
@@ -244,11 +244,11 @@ export var FeatureLayer = FeatureManager.extend({
    */
 
   eachActiveFeature: function (fn, context) {
-    // figure out roughly which layers are in view
+    // figure out (roughly) which layers are in view
     if (this._map) {
       var activeBounds = this._map.getBounds();
       for (var i in this._layers) {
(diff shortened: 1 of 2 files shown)


keithpower:
Thanks @jgravois , good catch on the duplicate keys.   Your check for duplicates wasn't catching them in my testing.  I used this instead and it appears to work.  Note, the this_cache was also getting duplicates

```
if (-1 === this._currentSnapshot.indexOf(id)) {
    this._currentSnapshot.push(id);
}
if (-1 === this._cache[key].indexOf(id)) {
    this._cache[key].push(id);
}
```

The eachActiveFeature is very similar to how I implemented it on the application level.  I will switch to this once I get time.  I used contains instead of intersect because I was only dealing with points.  I tihnk you'll want the check for the currentSnapshot in this function too.  If a _where_  is set at some point then this function will return features that may have been filtered out by the where.  So it could look like this instead.

```
 eachActiveFeature: function (fn, context) {
     // figure out roughly which layers are in view
     if (this._map) {
       var activeBounds = this._map.getBounds();
       for (var i in this._layers) {
         if (activeBounds.intersects(this._layers[i].getBounds()) &&  -1 !== this._currentSnapshot.indexOf(this._layers[i].feature.id)) {
           fn.call(context, this._layers[i]);
         }
       }
     }
     return this;
   },
```

I have some other potential improvements around the whole setWhere functionality that I will create a pull request for in the next week or so once I get past some deadlines.

jgravois:
@keithpower much obliged!

PR #941, opened 2017-03-28
Added ImageOveray Error handling.

Saulzi:
This builds on [leafet pr 5416](https://github.com/Leaflet/Leaflet/pull/5416) and adds basic error handling when loading of underlying ImageOverlays fails to esri-leaflet.

I am not sure how to add to your documentation with regards to adding the new event raised.

### DIFF
--- a/spec/Layers/RasterLayerSpec.js
+++ b/spec/Layers/RasterLayerSpec.js
@@ -6,4 +6,91 @@ describe('L.esri.RasterLayer', function () {
     layer._currentImage = null;
     expect(function () { layer.setOpacity(0.5); }).to.not.throw();
   });
+
+  // Extend 'abstract' RasterLayer object and implement functionality required for tests
+  var TestRasterLayer = L.esri.RasterLayer.extend({
+    initialize: function (options) {
+      this.service = {
+        metadata: function () {}
+      };
+    },
+
+    _buildExportParams: function () {
+      return 'data:image/gif;base64,R0lGODlhAQABAIAAAP///wAAACH5BAEAAAAALAAAAAABAAEAAAICRAEAOw=='; // A blank image URL
+    },
+
+    _requestExport: function (uri, bounds) {
+      this._renderImage(uri, bounds);
+    }
+  });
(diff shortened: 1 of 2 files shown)


jgravois:
so cool to see a user not only making contributions to _this_ project, but also upstream!

do you know of a way that i can simulate an error loading an image directly in the browser to check out the behavior in action?  i tried the below and it didn't throw

```js
// typo in service name ensures an error is returned from the server instead of an image
var dyn = L.esri.dynamicMapLayer({
  url:'http://sampleserver6.arcgisonline.com/arcgis/rest/services/Censs/MapServer',
}).addTo(map);

// not fired
dyn.on('error', function (e) {
  console.log(e);
})
```

Saulzi:
Have you made a build of leaflet as ImageLayer error functionality is 'bleeding edge' and was only merged into master yesterday. it should be in their next release. I will try your example shortly with a build of master.

Saulzi:
[Example.zip](https://github.com/Esri/esri-leaflet/files/879047/Example.zip)

The reason why your example was not working is because by default the layer will call the service which returns json with urls of images (this currently gives a different error above) which are then added to the map.

Setting the image type to gif basically short cuts this and will therefore raise the image.onerror event which is hooked up in the ImageOverlay (and subscribed to in DynamicMapLayer).

Please find attached copy of your example above with build of master on leaflet and the image-overlay-error-support branch from my repo.

jgravois:
> Have you made a build of leaflet [...]?

whoops! thanks for reminding me. 

a little local [`npm link`](https://docs.npmjs.com/cli/link) action (along with remembering to check out your own branch again) was all i needed to see your work for myself.  thanks again for this contribution.

seeing as how i just released `v2.0.8` i feel good about merging this work now and holding off on tagging again until _after_ leaflet `v1.1.0` lands.

as for the doc, i'll add it [here](https://github.com/Esri/esri-leaflet-doc/blob/master/src/pages/api-reference/layers/dynamic-map-layer.md#events) shortly.

PR #959, opened 2017-06-02
Added SetGeometry method

bbehling:
Copied the _setGeoemtry() method and import statements from the Query.js file.

Modified at() method so it uses the _setGeometry() method.

### DIFF
--- a/src/Tasks/IdentifyFeatures.js
+++ b/src/Tasks/IdentifyFeatures.js
@@ -1,10 +1,8 @@
-import { LatLngBounds, LatLng, GeoJSON, Utils } from 'leaflet';
+import { Utils } from 'leaflet';
 import { Identify } from './Identify';
-import { warn,
-  responseToFeatureCollection,
+import { responseToFeatureCollection,
   boundsToExtent,
-  geojsonToArcGIS,
-  geojsonTypeToArcGIS
+  setGeometry
 } from '../Util';
 
 export var IdentifyFeatures = Identify.extend({
@@ -31,7 +29,7 @@ export var IdentifyFeatures = Identify.extend({
   },
 
   at: function (geometry) {
-    Utils.setGeometry(geometry);
+    setGeometry(geometry);
     return this;
   },
 
--- a/src/Tasks/IdentifyFeatures.js
+++ b/src/Tasks/IdentifyFeatures.js
@@ -4,7 +4,7 @@ import { warn,
   responseToFeatureCollection,
   boundsToExtent,
   geojsonToArcGIS,
-  geojsonTypeToArcGIS 
+  geojsonTypeToArcGIS
 } from '../Util';
 
 export var IdentifyFeatures = Identify.extend({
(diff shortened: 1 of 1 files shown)


src/Tasks/IdentifyFeatures.js
```diff
@@ -62,6 +65,61 @@ export var IdentifyFeatures = Identify.extend({
         callback.call(context, undefined, featureCollection, response);
       }
     });
+  },
+  _setGeometry: function (geometry) {
```
jgravois: lets DRY this up and move  `_setGeometry()`   into `L.esri.Util` so that we can reference the same function from both `Query` _and_ `identifyFeatures`

jgravois:
the failing tests are the result of the fact that the signature of the requests we are sending to identify features has changed

```
// was
geometry=-122.66%2C45.51
// now
geometry={"x":-122.66,"y":45.51,"spatialReference":{"wkid":4326}}
```
since the new syntax [_is_ valid](http://resources.arcgis.com/en/help/arcgis-rest-api/index.html#/Identify_Map_Service/02r300000113000000/), we just need to update the tests. bonus points if you can spare the time to add a new test which passes through a bounding box or some other geometry type.

bbehling:
Module /home/travis/build/Esri/esri-leaflet/src/Util.js does not export setGeometry (imported by /home/travis/build/Esri/esri-leaflet/src/Tasks/IdentifyFeatures.js)

Is that error occurring due to the changed request signature?

bbehling:
Is there a need for a return statement at the end of the setGeometry() method?

bbehling:
IdentifyFeaturesSpec.js ->
expect(request.url).to.contain('geometry=-122.66%2C45.51');

We need to edit those asserts? Those are encoded, but your comment above says is not supposed to be encoded.

Let me know if that's the correct assert and whether the format is encoded or unencoded, and I'll make the change.

jgravois:
the asserts should remain encoded.  i (confusingly) decoded them when comparing because i thought it was more legible.

jgravois:
at this point, it'd be a good idea to close #960 and merge the commits in your `patch-2` branch into `patch-1`. i'm happy to help you untangle if you get stuck. otherwise i'll wait and code review when all your proposed changes are in one place.

bbehling:
Yes, please help untangle. I'm not sure which tools you want to use to merge to the changes, or if you want to do it manually. But please help. I haven't updated the specs yet.

jgravois:
> Yes, please help untangle.

no worries.  i pulled down both your branches and used plain old git to
1. create my own new branch using `bbehling:patch-1` as a base
2. merge `bbehling:patch-2` into that branch
3. then i made a few updates of my own
    1. refactored the code to streamline usage from two different classes
    2. made Util._setGeometry private (again)
    3. fixed the existing tests
4. afterward, i rebased my own branch (`jgravois:patch958` to squash some of your redundant commits down and opened up a new pull request to track from here

thanks for chipping in to help implement the new functionality @bbehling!  i'm going to be out of the office for a few days, but i'll be able to land #962 when i get back next week.

bbehling:
Thank you!

This is the first time I've used GITHub online editors. Wasn't sure how to merge here and what 'patches' are.

Issue #991, opened 2017-08-12
Metadata not fetched for tiledMapLayer with subdomains configured 


Biboba:
* Browser and version: 
Chromium Version 59.0.3071.109 (Build de développement) Built on Ubuntu , running on Ubuntu 16.04 (64 bits)

* Version of Leaflet (`L.version`): 1.2.0

* Version of esri Leaflet (`L.esri.VERSION`): 2.1.0

Steps to reproduce the error:

When using a L.esri.tiledMapLayer configured with subdomains (eg: https://{s}.arcgisonline.com/arcgis/rest/services/World_Topo_Map/MapServer, with ['server', 'services'] as subdomains),  
service metadata are not fetched as the url contains '{s}' notation throwing an "net::ERR_NAME_NOT_RESOLVED" error on Chromium Version 59.0.3071.109. 

Not critical though as the tiles get properly displayed.

On the following jsbin, I am using an arcgis online service as an example but I know I can use the 'basemapLayer' method which do not send this request.

http://jsbin.com/soropohemi/edit?html,output

jgravois:
thats a good catch. thank you for the thorough report! just to reiterate (and clarify)

### the bug

```js
var map = L.map("map").setView([37.75, -122.23], 10);
L.esri.tiledMapLayer({
  url: 'https://{s}.arcgisonline.com/arcgis/rest/services/World_Topo_Map/MapServer',
  subdomains: ['server', 'services']
}).addTo(map);
```

the code above *should* cause the copyright text printed [here](https://services.arcgisonline.com/arcgis/rest/services/World_Topo_Map/MapServer) to display in Leaflet's attribution control.
![screenshot 2017-08-12 21 50 09](https://user-images.githubusercontent.com/3011734/29246237-2c3a9a52-7fa8-11e7-9bbc-45e172a8a2cc.png)

this is what currently displays

![screenshot 2017-08-12 21 50 25](https://user-images.githubusercontent.com/3011734/29246240-3ea1646e-7fa8-11e7-9ce3-248e6960f81c.png)

if anyone is interested in helping patch this bug (i'm looking at you @daraobeirne 😺 ), its a pretty **fun** one!

[...]

PR #992, opened 2017-08-13
fix TiledMapLayer url with subdomains

Biboba:
Fix proposal for issue 991:
https://github.com/Esri/esri-leaflet/issues/991

### DIFF
--- a/src/Layers/TiledMapLayer.js
+++ b/src/Layers/TiledMapLayer.js
@@ -38,6 +38,11 @@ export var TiledMapLayer = TileLayer.extend({
   },
 
   initialize: function (options) {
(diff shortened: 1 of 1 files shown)


Biboba:
Wrong fix. Subdomains are not used anwmore. Will submit another one.

jgravois:
just a heads up for next time, whether you or a reviewer doesn't consider your initial proposal 'perfect', you can always continue refining *without* opening a totally separate pull request.

if you push additional commits to your fork they'll show up in the PR automatically.

ref: https://github.com/Esri/ago-assistant/pull/197

either way, thanks for pointing out the problem and digging in deeper to propose a fix all on your own!

PR #993, opened 2017-08-13
fix issue 991

Biboba:
Fix proposal for issue 991:
https://github.com/Esri/esri-leaflet/issues/991


jgravois:
very concise approach. nice work!

Issue #952, opened 2017-05-01
set f:'image' as the default for dynamic/imageMapLayer (again)


jgravois:
since we started defaulting to `f:'json'` in [`v1.0.0-rc.7`](https://github.com/Esri/esri-leaflet/releases/tag/v1.0.0-rc.7), quite a few folks have encountered errors that were resolved simply by fetching an output image instead of json.  this leads me to believe that it is a more appropriate default and a change we should investigate reverting.

#951, #862, #847, #782, #672 

granted, a handful of the above turned out to be bugs in our code that i've since fixed, but i'm still convinced that most folks would be better off fetching images most of the time.

i'll need to dig in a bit deeper to confirm there are no unexpected downstream consequences. my hunch is that Leaflet's own improved `imageOverlay` error handling (contributed by one of our users in https://github.com/Leaflet/Leaflet/pull/5416!) combined with the `crossOrigin` constructor option that landed long ago make it totally feasible for us to switch back.

jgravois:
granted the bar is pretty low right now, but i think this is more trouble than its worth.

jgravois:
i'm definitely still interested in reviewing if anyone wants to pursue this, but the related error reports have definitely died out.

PR #998, opened 2017-08-22
Fix content type for ImageMapLayer

nickpeihl:
The default server response content type per the [docs](http://esri.github.io/esri-leaflet/api-reference/layers/image-map-layer.html#options) for an `imageMapLayer` should be `image`, not `json`.

### DIFF
--- a/src/Layers/ImageMapLayer.js
+++ b/src/Layers/ImageMapLayer.js
@@ -9,7 +9,7 @@ export var ImageMapLayer = RasterLayer.extend({
     updateInterval: 150,
     format: 'jpgpng',
     transparent: true,
-    f: 'json'
+    f: 'image'
   },
 
   query: function () {


nickpeihl:
[Here's a working example](http://jsbin.com/rowemuhosa/edit?html,output) where the default `f: 'json'` parameter is superceded by `f: 'image'`.

jgravois:
thanks for this @nickpeihl!

you have any interest in helping me reopen the can of worms in https://github.com/Esri/esri-leaflet/issues/952?

nickpeihl:
I should have searched before submitting the PR. I didn't know it was previously discussed. Let me dig in a bit and see where I can help.

jgravois:
no need to apologize. having 'imageMapLayer' fetching raw images is an improvement in my book and a good first step toward doing the same with 'dynamicMapLayer'.

i seem to remember the other change being error inducing in the test suite, but between the two of us, i'm sure we'll be able to figure it out.

Issue #1001, opened 2017-08-31
Create returnM method for setting param on Query task


jfolds:
L.esri.Query does not have a returnM method. It currently exists in the Find task, but is missing from Query.

jgravois:
you up for contributing it as a new feature? i'd be happy to review a PR.

https://github.com/Esri/esri-leaflet/blob/58f05b33ca0b510c366e5cd1a6b4031accfee683/src/Tasks/Query.js#L11-L21

jfolds:
Sure, I'll try and submit a PR soon.

jgravois:
thank you @jfolds! in the meantime, folks can also use this 'weird trick'

```js
var query = L.esri.query({
    url:'https://services.arcgis.com/.../stops/FeatureServer/0'
});

query.within(bounds);

query.params.returnM = true;

query.run(function(error, featureCollection, response){
    console.log('Found ' + featureCollection.features.length + ' features');
});
```

PR #1002, opened 2017-08-31
add returnM setter to Query task

jfolds:
addresses #1001


jgravois:
i pushed another commit with a test. thank you @jfolds!

PR #1011, opened 2017-10-06
dynamically resample Esri World Imagery

### LINKED ISSUE #240: Latest ArcGis jsapi library resamples images, this doesn't
Latest versions of the ArcGis jsapi seem to be able to resample tiles at high zoom levels to prevent the annoying "no data available" images being shown to users if they are not available; but this library doesn't seem to have this capability. Are there any plans to add this feature? I'm not actually sure how the ArcGis library detects when it's received a "no data" image, so I've no idea how hard this might be.

best

jgravois:
resolves #240 (better late than never!)

instead of artificially capping the zoom for World Imagery at `19`, we can interrogate an esri tilemap to see whether all tiles in an 8x8 grid at the destination zoom level are present.

if any tiles are missing, Leaflet has a nifty built-in option to help us downsample tiles from a level further out.

### to do
- [ ] figure out whether `zoomanim` fires on pinch zooms
- [ ] fetch a tilemap on pan too?
- [ ] avoid the screen flash when missing tiles are replaced with resampled tiles?

tilemap

### DIFF
--- a/src/Layers/BasemapLayer.js
+++ b/src/Layers/BasemapLayer.js
@@ -102,7 +102,9 @@ export var BasemapLayer = TileLayer.extend({
         urlTemplate: tileProtocol + '//{s}.arcgisonline.com/ArcGIS/rest/services/World_Imagery/MapServer/tile/{z}/{y}/{x}',
         options: {
           minZoom: 1,
-          maxZoom: 19,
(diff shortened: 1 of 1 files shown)


jgravois:
yolo

PR #1015, opened 2017-10-19
Add jsDelivr hits badge

LukasDrgon:
Hey, I noticed that esri-leaflet gets 1M hits per month on jsDelivr, so I thought you might like this badge.


jgravois:
very cool. thank you!

we've had +3 million hits on unpkg.com in the last 30 days too!

PR #1021, opened 2017-11-04
fix setLayers() : issue 1012

Biboba:


src/Layers/DynamicMapLayer.js
```diff
@@ -42,6 +42,12 @@ export var DynamicMapLayer = RasterLayer.extend({
   },
 
   setLayers: function (layers) {
+    if (!layers) { layers = []; }
+    if (layers.length === 0 && this.options.layers.length) {
```
jgravois: the second check is problematic when a dynamicMapLayer is instantiated _without_ a layers constructor parameter and `.setLayers([])` is called later.

Biboba:
Hi @jgravois ,

Thanks for the feedback. Indeed, this solution is not ideal. The problem you described is only appearing when the extent of the map has not changed [source code](https://github.com/Esri/esri-leaflet/blob/808e67d215eef721724a9027df60b5c69cc751e5/src/Layers/RasterLayer.js#L40).

Your solution seems simpler and working better !
I just made a small fix in [c08f89](https://github.com/Biboba/esri-leaflet/commit/c08f89e38c649d82e12eeaa9cac428846400aad8) that was breaking paning on the map once no layers had been configured.

Thanks

jgravois:
> I just made a small fix in `c08f89` that was breaking panning on the map once no layers had been configured.

very cool! if you can get this branch in sync, I'll merge your PR. thanks again for the help!

Biboba:
Done. 
Branch "issue-1012" on my repo is now up to date ([commit](https://github.com/Biboba/esri-leaflet/commit/19f8b691979fbcca40beff91a972153e663c98b0)).
Thanks

src/Layers/DynamicMapLayer.js
```diff
@@ -42,6 +42,12 @@ export var DynamicMapLayer = RasterLayer.extend({
...
+      this.onRemove(this._map);
+    } else if (layers.length && this.options.layers.length === 0) {
+      this.onAdd(this._map);
+    }
     this.options.layers = layers;
     this._update();
```
jgravois: now the ._update() call throws `Uncaught TypeError: Cannot read property 'removeLayer' of null` in the following situation.

```js
.setLayers([1]);
// wait
.setLayers([]);
```
hopefully its just an order of operations thing.

PR #1025, opened 2017-11-22
use Chrome for CI tests instead of phantomjs

### LINKED ISSUE #907: upgrade to PhantomJS 2.x
see #896 for more info

### LINKED ISSUE #967: test failures with leaflet 1.1.0
https://travis-ci.org/Esri/esri-leaflet/builds/243458201

we're seeing a handful of different failures, but they all result from `FeatureLayer.createFeatures()` not being called. manual testing confirms they are false positives, but i haven't been able to figure out why the spies aren't being triggered.

_typically_ rogue errors like this are caused by the fact that the actual requests don't look enough like the urls sinon.fakeServer is poised to send responses for.

with this in mind, i tried substituting an overly generic regex, but found no joy.
 
```js
server.respondWith('GET', new RegExp(/http:\/\/gis.example.com\/mock\/arcgis\/rest\/services\/MockService\/MockFeatureServer\/0\/query\?.+/), JSON.stringify({
    fields: fields,
    features: [feature1, feature2],
    objectIdFieldName: 'OBJECTID'
}));
```

jgravois:
resolves #907 (by omission)

i've long been unable to figure out how to migrate to phantomjs 2.x
without breaking a significant number of tests.

on the contrary, only one real test was sacrificed to take this approach.
the added benefit is that it also resolves #967

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,8 +7,13 @@ node_js:
 cache:
   directories: # Cache dependencies
     - node_modules
-    - travis_phantomjs
 before_script:
   - npm prune
+  - "export DISPLAY=:99.0"
+  - "sh -e /etc/init.d/xvfb start"
+  - sleep 3 # give xvfb some time to start
 before_install:
   - npm i -g npm@latest
+  - google-chrome-stable --headless --disable-gpu
+addons:
+  chrome: stable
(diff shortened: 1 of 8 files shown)


patrickarlt:
@jgravois LGTM :shipit:

jgravois:
thank you!

PR #1029, opened 2017-12-04
remove injected script and style tags when map is destroyed

### LINKED ISSUE #1018: Cleanup injected attribution <style> and <script> tags
<style> and <script> tags are injected into the page for attribution reasons when adding basemap layers.

There should probably be a way to either:
a) prevent injecting <script> or <style> tags into the page
or
b) remove the injected tags when a map is destroyed

jfolds:
resolves #1018 - tried to keep support to IE8+
let me know if any issues, thanks!


jgravois:
not sure what's up with Travis, but all tests are passing locally and your logic seems fine to me.

thanks for this @jfolds!

Issue #633, opened 2015-09-11
Should be able to pass identify options into bindPopup for a dynamic layer


anatolyg:
Right now there are defaults that are applied to the identify call from a dynamic layer. It would be great to be able to pass an options object that overwrote some of those. For example, I might not want to return the geometry for the feature, or, I might want to make the tolerance higher (or lower).

jgravois:
we typically expect people to call the [available methods](http://esri.github.io/esri-leaflet/api-reference/tasks/identify-features.html) for setting identify request properties in advance of calling `run()`.

```js
L.esri.Tasks.identifyFeatures({
    url: 'http://sampleserver6.arcgisonline.com/arcgis/rest/services/WorldTimeZones/MapServer'
})
.on(map)
.at([45.543, -122.621])
.layers('visible:1')
.returnGeometry(false)
.tolerance(3)
.run(function(error, featureCollection, response){
    console.log("UTC Offset: " + featureCollection.features[0].properties.ZONE);
});
```
is there a reason why this isn't ideal for you?

anatolyg:
I create the layer using `L.esri.dynamicMapLayer` and then use the `bindPopup` to set up popup. I may do this to multiple layers on the same map, so it's really convenient that the layer itself handles the identify service creation (via `_getPopupData` function). But I don't see a way to pass options to the identify task from instanciating the layer this way. Totally possible I am missing something.

jgravois:
thanks for clarifying..  i'll have to think about this a bit to try and decide the best way to expose an opportunity to get more fine grained control (ie: dynamicMapService initialization, when calling `bindPopup()` etc.)

would it be sufficient to be able to pass identifyParameters that applied to *all* layers, or are you expecting to set things differently layer by layer?

anatolyg:
I hope layer by layer. Use case is time-based identify calls -- maybe you want to see data for a layer a year in the future of data for another layer. The single parameter set won't do there. But bigger picture, it would be useful to set a variety of parameters that are currently exposed via methods as parameters into the `identifyFeatures` call (and then allow for an `identify` field to be set during `dynamicMapLayer` init).

jgravois:
but for the time being you can just listen for the map click, fire your own identifyFeatures request and display the results in a popup, correct?

anatolyg:
Oh, for sure! This isn't blocking me, just a nice to have. I really like the bindPopup feature, just thought you guys might be interested in some feedback.

jgravois:
[...]

PR #1031, opened 2017-12-11
Add popup options to DynamicMapLayer

Biboba:
Add through a "popup" option to the DynamicMapLayer constructor, the ability to pass identify options into bindPopup (issue [633](https://github.com/Esri/esri-leaflet/issues/633))

### DIFF
--- a/src/Layers/DynamicMapLayer.js
+++ b/src/Layers/DynamicMapLayer.js
@@ -87,19 +87,26 @@ export var DynamicMapLayer = RasterLayer.extend({
       }, this), 300);
     }, this);
 
-    var identifyRequest = this.identify().on(this._map).at(e.latlng);
-
-    // remove extraneous vertices from response features
-    identifyRequest.simplify(this._map, 0.5);
-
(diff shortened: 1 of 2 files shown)


src/Layers/DynamicMapLayer.js
```diff
@@ -87,19 +87,26 @@ export var DynamicMapLayer = RasterLayer.extend({
       }, this), 300);
     }, this);
 
-    var identifyRequest = this.identify().on(this._map).at(e.latlng);
+    var identifyRequest;
+    if (this.options.popup) {
+      identifyRequest = this.options.popup.on(this._map).at(e.latlng);
```
jgravois: i pushed a small change here to call `IdentifyFeatures.on()` dynamically even when a custom popup constructor option has been bound to the layer. this is beneficial for two reasons.

1. it avoids an error when `on()` wasn't chained prior.
2. even if the method was chained, we need to avoid passing a stale extent through in the request if the map was panned or zoomed after the layer was instantiated.

PR #1035, opened 2017-12-30
fix TiledMapLayer metadata test + test for issue 991

Biboba:
- Fix 'should display a metadata attribution if one is present and no attribution option was passed' test for which assertion was not tested.
- Test for issue 991


jgravois:
thx @Biboba!

PR #1037, opened 2018-01-03
add test for issue633

Biboba:


jgravois:
thank you for this @Biboba! I _really_ appreciate contributions like this!

PR #1045, opened 2018-02-07
make check for custom CRS smarterer

### LINKED ISSUE #1019: proj4 is not defined in TiledMapLayer
* Browser and version:
Any browser you like
1.2.0
2.1.1
* Steps to reproduce the error:
1. Use `Leaflet` and `esri-leaflet` together with `proj4leaflet`
2. `proj4leaflet` has a dependency of `proj4`
3. Import `proj4leaflet` AMD style without importing `proj4` manually anywhere else since `proj4leaflet` will pick it up from dependencies and will work fine

What happens is that when doing this I will get an error on this line https://github.com/Esri/esri-leaflet/blob/master/src/Layers/TiledMapLayer.js#L141 saying `Uncaught ReferenceError: proj4 is not defined`.

I was expecting for this to work fine, since only the `proj4leaflet` library is dependant on `proj4` and `esri-leaflet` is not.

It is quite hard to replicate this problem using `jsbin`, since I use `npm` and `package.json` and `webpack`.

* Optional: I'm *not* using the [CDN](http://www.jsdelivr.com/projects/leaflet.esri), I'm loading/bundling the library using:
Webpack

Adding this as a workaround helps to get it work and no errors are shown, but I think that should not be required for `esri-leaflet` to work without issues.
```
import * as proj4 from 'proj4';
window['proj4'] = proj4;
```

jgravois:
resolves #1019 

took a look at this today and realized that the problem @theashyster reported also causes an unnecessary console error _outside_ of webpack when a simple WGS84 tile service is added to a map.

```js
var map = L.map('map',{crs: L.CRS.EPSG4326}).setView([45,-125], 3);

L.esri.tiledMapLayer({
  url: 'http://services.arcgisonline.com/arcgis/rest/services/ESRI_Imagery_World_2D/MapServer'
}).addTo(map);
```
instead of checking to see whether a) the map and tiled service are both web mercator and b) if `proj4` is available we can instrospect the CRS defined for the map and make sure the supplied code is the same one the service is using.

### DIFF
--- a/src/Layers/TiledMapLayer.js
+++ b/src/Layers/TiledMapLayer.js
@@ -118,7 +118,10 @@ export var TiledMapLayer = TileLayer.extend({
             this.options.attribution = metadata.copyrightText;
             map.attributionControl.addAttribution(this.getAttribution());
           }
-          if (map.options.crs === L.CRS.EPSG3857 && (sr === 102100 || sr === 3857)) {
+
+          var needProj4 = this._checkSRSupport(map, sr);
+
+          if (needProj4) {
             this._lodMap = {};
             // create the zoom level data
             var arcgisLODs = metadata.tileInfo.lods;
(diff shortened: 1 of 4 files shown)


src/Layers/TiledMapLayer.js
```diff
@@ -149,6 +150,22 @@ export var TiledMapLayer = TileLayer.extend({
...
+  _checkSRSupport: function (map, spatialReference) {
+    if (map.options.crs === L.CRS.EPSG3857 && (spatialReference === 102100 || spatialReference === 3857)) {
+      // web mercator tile services are supported by leaflet
+      return true;
+    } else if (map.options.crs === L.CRS.EPSG4326 && (spatialReference === 4326)) {
+      // wgs84 tile services are supported by leaflet
+      return true;
+    } else if (map.options.crs && map.options.crs.code && (map.options.crs.code.indexOf(spatialReference) > -1)) {
+      // using proj4 and defining a custom crs enables support for tile services published in other projections
+      return true;
```
theashyster: Hey, today I found the time to test out the newest version of `esri-leaflet` and sadly it did not work out as I hoped for.
Before the changes things related to `proj4` would go into the `else` block and show a warning if `proj4` was missing (which caused the original issue I described here https://github.com/Esri/esri-leaflet/issues/1019).
After the changes things related to `proj4` goes into the main block and fires `lodmap` event.
Because of this, I only get the initial map loaded and when I do zooming or dragging around the map no other base map tiles are loaded or even requested to be loaded.

Here is the issue visible on Plunker https://plnkr.co/edit/kFneHCA8CBMpAbg12mwp?p=preview
And if you just change the version to `esri-leaflet@2.1.2` in the `index.html` you would see that it works fine.

jgravois: ugh. totally my fault. fix is in #1051.

theashyster: Heh, no problem, thanks for the fast fix 👍

PR #1049, opened 2018-02-12
Convert to date on arcgis to geojson

Saulzi:
Please find attached pull request which changes the current esri-geojson functionality.

At present date fields from arcgis-server are not converted to something representing a date in the geojson (often iso8601) which is used for Feature layers. Date fields are just time since epoch and there is currently no way to get the esri field type once the conversion from esri > geojson has been done. 

This pull request changes current functionality to convert dates to javascript dates in the geo-json for features, this may or may not be the correct approach.

It is enougth to meet my requirements but may break may or may not break the filtering dates on clientside related stuff in featuremanager (I am not using this functionality), i feel that this may need a bunch of tests around both if fields are specified and not in data from arcgis-server to see what happens here.

another solution may be to append the fields property to each geojson feature and then use this to convert date fields to required type in my own code,

### DIFF
--- a/spec/Layers/FeatureLayer/FeatureManagerSpec.js
+++ b/spec/Layers/FeatureLayer/FeatureManagerSpec.js
@@ -257,7 +257,7 @@ describe('L.esri.FeatureManager', function () {
           'OBJECTID': 2,
           'Name': 'Site 2',
           'Type': 'Inactive',
-          'Time': new Date('January 15 2014 GMT-0800').valueOf()
+          'Time': new Date('January 15 2014 GMT-0800')
         },
         'id': 2
       }
@@ -293,7 +293,7 @@ describe('L.esri.FeatureManager', function () {
           'OBJECTID': 2,
           'Name': 'Site 2',
           'Type': 'Inactive',
-          'Time': new Date('January 15 2014 GMT-0800').valueOf()
+          'Time': new Date('January 15 2014 GMT-0800')
         },
         'id': 2
       },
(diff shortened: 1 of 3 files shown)


jgravois:
thx for this @Saulzi!

to be honest i'm on the fence about whether its our place to do this work internally whenever we have a fieldmap to work with (when using `Find` or `Identify` we don't) or not.

> another solution may be to append the fields property to each geojson feature 

i don't love the idea of hanging this info off of each individual feature, but i'm fairly certain that we could come up with something else that'd be equally useful.

How are you retrieving your geojson? `FeatureLayer`? `Query`?

either way, give me some time to think about this. i'll likely come around.

Saulzi:
I thought that that would probably be the the case, 

I am using FeatureLayer rather than Query, I am also not keen on hanging it off each feature, allthough its only a reference so should not have too much impact, 

Is this metadata likely to change for each request, allthough it seems that it is actually there on the initial call to mapserver/?f=json (on the version of ArcGIS Server i'm talking to). This is done in in FeatureManager.OnAdd. Perhaps it could be exposed on the layer (Extends FeatureManager), If so i would be able to use the metadata when I am dealing with the arcgis features which have been converted to geojson to know that the integer is actually a date.

jgravois:
> Is this metadata likely to change for each request

the schema of the layer within the service will not change unless the service itself is overwritten. this makes your suggestion of exposing it on `FeatureLayer` itself using the existing metadata check that fires when the layer is added to the map a superb choice.

Saulzi:
Looking into this a little bit further the FeatureManager allready retrieves and caches the Metadata when there is a call to the private _getMetadata function, there are a few other calls which use this.service.metadata which do not cache the previous result, is this intentional?

My thoughts are that all of the calls in this class should be brought in line so that they use the private _getMetadata function, as as previously mentioned the metadata should be pretty static.

I am using the metadata function to retrieve the metadata, which is good enougth for now but adds additional chatter on an allready chatty library,

jgravois:
> there are a few other calls which use this.service.metadata which do not cache the previous result, is this intentional?

definitely not.

jgravois:
closing, but more than happy to discuss further if anyone has a question/comment down the line.

PR #1051, opened 2018-02-15
fix logic for determining when to map lods

jgravois:
i made a serious logic error in #1045

we should only be attempting to remap LODs when the CRS of the map is web mercator and the LODs roughly match the convention

for WGS84 tiled services and applications with an appropriately defined custom CRS (using Proj4) that matches the service, we should do nothing.

we should only log a warning to the console when neither of the above are true.

cc/ @theashyster

### DIFF
--- a/src/Layers/TiledMapLayer.js
+++ b/src/Layers/TiledMapLayer.js
@@ -114,14 +114,14 @@ export var TiledMapLayer = TileLayer.extend({
       this.metadata(function (error, metadata) {
         if (!error && metadata.spatialReference) {
           var sr = metadata.spatialReference.latestWkid || metadata.spatialReference.wkid;
+          // display the copyright text from the service using leaflet's attribution control
           if (!this.options.attribution && map.attributionControl && metadata.copyrightText) {
             this.options.attribution = metadata.copyrightText;
             map.attributionControl.addAttribution(this.getAttribution());
           }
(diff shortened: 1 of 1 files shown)


src/Layers/TiledMapLayer.js
```diff
@@ -140,7 +140,10 @@ export var TiledMapLayer = TileLayer.extend({
             }
 
             this.fire('lodmap');
+          } else if (map.options.crs && map.options.crs.code && (map.options.crs.code.indexOf(sr) > -1) || (map.options.crs === CRS.EPSG4326 && (sr === 4326))) {
```
theashyster: Hmm, I think the second part is actually redundant here.
I looked at `CRS.EPSG4326` here https://github.com/Leaflet/Leaflet/blob/master/src/geo/crs/CRS.EPSG4326.js#L20 and it also contains `code` attribute, so the code would go in the `else if` block already after the first check that is `map.options.crs && map.options.crs.code && (map.options.crs.code.indexOf(sr) > -1)`

So at the end it looks to me that this `else if` block could be simplified to this
`} else if (map.options.crs && map.options.crs.code && map.options.crs.code.indexOf(sr) > -1) {`

jgravois: good catch!

PR #1057, opened 2018-03-06
update all tests with fetch-mock library -wip

Biboba:
Update tests with [fetch-mock](https://www.npmjs.com/package/fetch-mock) library.
Still some tests are failing:

- **DynamicMapLayer, ImageMapLayer:**
3 tests failing because of a problem popup related:
•	Can’t find a way to do the test when the popup has rendered
•	fetchMock.flush() does not help : promise has resolved does not mean that callback has been executed

- **DynamicMapLayer, ImageMapLayer, TiledMapSpec:** 
Test "_should propagate events from the service_" failing:
Don't know how to handle spy on requestend because as the event is triggered after the metadata callback has been called, I can’t find a way to « hook » the test: `expect(requestendSpy.callCount).to.be.above(0);`

- **FindSpec, IdentifyImageSpec, QuerySpec:**
Part of the tests testing for xhr type are commented : should the promise be returned ? So far it is not the case.
`expect(this.request).to.be.an.instanceof(XMLHttpRequest);`

- **RequestSpec:**
2 tests testing for error failing : failed not managed yet [code]https://github.com/Esri/esri-leaflet/blob/51940752bd3feee6ad53eb146d9cc2ae030d77d6/src/Request.js#L8)

- [ ] L.esri.FeatureManager test TODO.
@jgravois I wanted to have your opinion so far on the issues above before going ahead as at a first look it seemed trickier for FeatureManager

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -14,6 +14,7 @@ module.exports = function (config) {
     // list of files / patterns to load in the browser
     // not sure why tests are failing when files are loaded in bulk
     files: [
+      'node_modules/fetch-mock/es5/client-bundle.js',
       'node_modules/leaflet/dist/leaflet.css',
       'node_modules/leaflet/dist/leaflet-src.js',
       'dist/esri-leaflet-debug.js',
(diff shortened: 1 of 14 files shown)


jgravois:
wow! heroic work @Biboba! it'll probably be next week before i have time to dig in further, but this is _great_ progress.

Issue #1069, opened 2018-03-13
No callback for useCors=false and Server Error


strajuser:
There is no callback for request with useCors=false and any server error.
As example, any not-found-url:
http://jsbin.com/nesomeyiju/edit?html,console,output

Can be simply fixed by addind script.onerror handler in src/Requests.js, e.g:
script.onerror = function(error) { callback.call(context, error); }

jgravois:
thx for taking the time to point this out @strajuser! a pull request would be sincerely appreciated. if that's not feasible, i will patch myself in time.

strajuser:
#1070 =)

jgravois:
wow. its been that kind of week. thanks again!

PR #1070, opened 2018-03-13
no-callback fix for jsonp with server error

strajuser:
Fix for no-callback jsonp request with server error [https://github.com/Esri/esri-leaflet/issues/1069](#1069) with onerror handler.

### DIFF
--- a/src/Request.js
+++ b/src/Request.js
@@ -187,6 +187,20 @@ export function jsonp (url, params, callback, context) {
   script.type = 'text/javascript';
   script.src = url + '?' + serialize(params);
   script.id = callbackId;
+  script.onerror = function (error) {
+    if (error && window._EsriLeafletCallbacks[callbackId] !== true) {
+      // Can't get true error code: it can be 404, or 401, or 500
+      var err = {
+        error: {
+          code: 500,
+          message: 'Server error'
+        }
+      };
+
+      callback.call(context, err);
+      window._EsriLeafletCallbacks[callbackId] = true;
+    }
+  };
   DomUtil.addClass(script, 'esri-leaflet-jsonp');
 
   callbacks++;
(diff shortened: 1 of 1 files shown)


strajuser:
All local tests was Ok. 
Sorry about bad username in commit, I can change it in another PR.

src/Request.js
```diff
@@ -187,6 +187,20 @@ export function jsonp (url, params, callback, context) {
   script.type = 'text/javascript';
   script.src = url + '?' + serialize(params);
   script.id = callbackId;
+  script.onerror = function (error) {
+    if (error && window._EsriLeafletCallbacks[callbackId] !== true) {
+      // Can't get true error code: it can be 404, or 401, or 500
+      var err = {
+        error: {
+          code: 500,
```
jgravois: if we don't know what the actual error was, best not to assign a code arbitrarily.

what about something like this?
```js
var err = {
  message: 'An unknown error occurred'
}
```

strajuser: Ok, it makes sense. I've just tried to save error contract as
```javascript
var err = {
  error: {
     code: 'need something undefined',
     message: 'An unknown error occurred'
  }
}
```

May be close this PR and make another with correct username and message fix?
And what about failed tests?

jgravois: > I've just tried to save error contract 

meh, [this guy](https://stackoverflow.com/questions/5334413/parsing-jsonp-response-in-javascript-when-4xx-or-5xx-http-error-code-is-present) seems to agree with you, so i can live with a 500 if you update the message. 

> May be close this PR and make another with correct username

no need to close this PR. just push another commit to the same branch (and rebase if it'd make you happy).

> what about failed tests?

those weren't your fault. i restarted the build and am seeing 💚 now.

PR #1076, opened 2018-03-28
Added missing attribution option

octavm:
Added missing attribtion option to OceansLabels and ImageryTransportation basemaps. 
Because of null attribution following error was happening: 
Uncaught TypeError: Cannot set property 'innerHTML' of null
    at _updateMapAttribution (Util.js:342)
    at Util.js:313

### DIFF
--- a/src/Layers/BasemapLayer.js
+++ b/src/Layers/BasemapLayer.js
@@ -47,7 +47,8 @@ export var BasemapLayer = TileLayer.extend({
           minZoom: 1,
           maxZoom: 16,
           subdomains: ['server', 'services'],
-          pane: (pointerEvents) ? 'esri-labels' : 'tilePane'
+          pane: (pointerEvents) ? 'esri-labels' : 'tilePane',
+          attribution: ''
         }
       },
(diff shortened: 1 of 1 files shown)


jgravois:
i don't understand what problem this pull request solves.

octavm:
What do you mean? I just put in the description that the " Cannot set property 'innerHTML' of null" error happens because 'attribution' option is missing for the two basemaps.

The simplest of test cases can replicate that error. 
Adding the 'OceansLabels' basemap, and then dragging a few times, makes the map freeze. The error is logged in the console.
Its really not that hard to replicate it and see the error..

jgravois:
> The simplest of test cases can replicate it. Here, I've created this [CodePen](https://codepen.io/anon/pen/WzMXQw) to see the error.

I took a look at your CodePen and it didn't freeze or throw errors in the console after panning/zooming.

jgravois:
even if i can't reproduce the error, its a harmless change. thanks @octavm.

octavm:
My pleasure to help.

PR #1078, opened 2018-03-28
Check if 'attributionElement' exists. Fixes #1077

### LINKED ISSUE #1077: Error :"Cannot set property 'innerHTML' of null" when attribution is missing
I'm using Chrome 65.0.3325.181 (Official Build) (64-bit), leaflet-1.3.1, esri-leaflet-2.1.4 master, 

Steps to reproduce the error:

1. Set the attribution option of TileLayer to 'null' for any Esri BasempLayer
2. Drag the map a few times, zoom in, out. After a few moves, it will freeze.
3. See F12 for errors:
`Uncaught TypeError: Cannot set property 'innerHTML' of null
    at M (Util.js:342)
    at Util.js:313
    at Object.window._EsriLeafletCallbacks.(anonymous function) [as c0] (https://unpkg.com/esri-leaflet@2.1.3/dist/esri-leaflet.js:4:2566)
    at World_Street_Map?callback=window._EsriLeafletCallbacks.c0&f=json:1`

Here is a CodePen that demonstrates the problem: ([CodePen](https://codepen.io/anon/pen/wmpQPo))

This means that whenever the attribution is null or missing completely, BasemapLayer don't work anymore.
Also I've seen problems with the "*Labels" basemaps, that have attribution=''. When Debugging the code the same error happens.

octavm:
When doing _updateMapAttribution, make sure to check if attributionElement actually exists to avoid "TypeError: Cannot set property 'innerHTML' of null". Fixes #1077

### DIFF
--- a/src/Util.js
+++ b/src/Util.js
@@ -317,8 +317,9 @@ export function _getAttributionData (url, map) {
 export function _updateMapAttribution (evt) {
   var map = evt.target;
   var oldAttributions = map._esriAttributions;
+  var attributionElement = map.attributionControl._container.querySelector('.esri-dynamic-attribution');
 
-  if (map && map.attributionControl && oldAttributions) {
+  if (map && map.attributionControl && attributionElement && oldAttributions) {
     var newAttributions = '';
     var bounds = map.getBounds();
     var wrappedBounds = latLngBounds(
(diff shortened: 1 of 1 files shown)


jgravois:
i'm still not sure why a developer would explicitly nullify the attribution for our basemap specifically to avoid giving credit where credit is due, but there is no harm in trapping for it.

thanks @octavm.

octavm:
Glad I could help :-)

Issue #1055, opened 2018-03-01
Editing multiple features at once


mladenp:
It looks like methods for editing multiple features doesn't exist in this library.
I can only find addFeature/updateFeature methods but they accept only one feature.
Currently i am calling ArcGIS updateFeatures manually, without esri-leaflet lib, by sending requests via form-data.
And i am sending features data like so: 
`[{"attributes":{"OBJECTID":2406107,"PROGRESS_STATUS":"Assigned"}}]`

So my question is: is there any way to edit multiple features at once and can i send just data i want to update, like i am already doing in my form-data request?

jgravois:
i added a method to bulk `deleteFeatures()` in #510. a pull request for something similar to `updateFeatures()` would be welcome.

Biboba:
Hi,

Would see an `addFeatures: function (feature, callback, context)` function, taking as input a feature GeoJSON FeatureCollection or an array of features or should support both ?

jgravois:
sorry, i thought i already replied here @Biboba. support for FeatureCollections would be sufficient.

jgravois:
resolved (way back in April) in #1083

PR #1083, opened 2018-04-07
add and update several features in FeatureLayer

Biboba:
Add the possibility to add and update several features (`updateFeatures`, `addFeatures`) by specifying a geojson featureCollection object ([issue 1055](https://github.com/Esri/esri-leaflet/issues/1055))

### DIFF
--- a/spec/Layers/FeatureLayer/FeatureManagerSpec.js
+++ b/spec/Layers/FeatureLayer/FeatureManagerSpec.js
@@ -759,43 +759,43 @@ describe('L.esri.FeatureManager', function () {
   it('should wrap the updateFeatures method on the underlying service and refresh', function (done) {
     server.respondWith('POST', 'http://gis.example.com/mock/arcgis/rest/services/MockService/MockFeatureServer/0/updateFeatures', JSON.stringify({
       'updateResults': [{
-        'objectId': 1,
+        'objectid': 1,
         'success': true
       }, {
-        'objectId': 2,
+        'objectid': 2,
         'success': true
       }]
     }));
 
     layer.updateFeatures({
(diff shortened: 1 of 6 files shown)


Biboba:
Though I think it would make more sense to have one single function `deleteFeatures`, `updateFeatures`, `addFeatures` handling any geojson (feature and featureCollection) just like the REST API does.

jgravois:
very cool @Biboba. thank you!

> I think it would make more sense to have one single function (...) handling any GeoJSON (feature and featureCollection)

that'd be okay with me as long as we keep the historic methods around.

Biboba:
So I left `addFeature`, `updateFeature`, `removeFeature` and respected the old response to the callback signature with a single element rather than an array.

Biboba:
Hi @jgravois,
Looks good to you ?
Thanks !

spec/Layers/FeatureLayer/FeatureManagerSpec.js
```diff
@@ -756,6 +756,54 @@ describe('L.esri.FeatureManager', function () {
...
+        properties: {
+          foo: 'bar'
+        },
+        geometry: {
+          type: 'Point',
+          coordinates: [-121, 45]
+        }
+      }]
+    }, function (error, response) {
+      expect(response).to.deep.equal([{
```
jgravois: i noticed that the syntax errors i fixed in 872de35 didn't actually induce an error in your test because you're currently only introspecting the mock service response (and not the actual _request_ parameters) which this lib actually has control over.

Biboba:
Great !
Thanks for the feedback @jgravois

Issue #1038, opened 2018-01-03
intermittent test failures on Travis


jgravois:
we are _still_ experiencing intermittent errors when running the test suite on [Travis](https://travis-ci.org/Esri/esri-leaflet).

i've had _some_ success in the past playing with the order individual tests are executed, but i'd love to actually put the problem to bed once and for all.

https://github.com/Esri/esri-leaflet/blob/a95cdce0b47bafc8716aeb08bfd78895a3c7f5c0/karma.conf.js#L14-L22

do you have any ideas @Biboba?

Biboba:
I do experienced it locally as well.
I am under the impression that some tests are failing depending on asynchronous timing.

I noticed that for example, 'it will load a new image when the map moves'  test of DynamicMapLayerSpec is falling quite often. 

https://github.com/Esri/esri-leaflet/blob/a95cdce0b47bafc8716aeb08bfd78895a3c7f5c0/spec/Layers/DynamicMapLayerSpec.js#L106-L122

In this one, there are nested layer load event, and by debugging it seems that when failing, it exits the test and fails before having "reach" the 'done' callback. 

<img width="485" alt="capture d ecran 2018-01-04 a 19 09 20" src="https://user-images.githubusercontent.com/11556423/34578273-9c8d0198-f184-11e7-8801-700073804abd.png">

jgravois:
> I am under the impression that some tests are failing depending on asynchronous timing.

what would be the best improvement we could make?

jgravois:
the tweaks in 043d19e912ca41c513096185cfd09828fe640eb6 seem to generate _slightly_ more predictable behavior, but we'll need to rethink how we set up and tear down the map in the DynamicMapLayer and ImageMapLayer tests in order to get to the **root** of the problem.

patrickarlt:
In general almost every travis build (except gh-release branch which should fail) has passed in the last few years so I'm going to close this https://travis-ci.org/Esri/esri-leaflet/builds.

jgravois:
I'm okay with closing this issue, but just as a heads up, besides test failures I've also seen a substantial number of runs where the test suite exited with a `0` after only executing a handful of the tests.

PR #1084, opened 2018-04-09
Add setZIndex function

appleshowc:
增加了手动设置和自动设置图层 zIndex 值的方法。
在初始化设置了 position 参数时，自动设置图层 zIndex 值（遍历所在 pane 里的所有图层，设置最大（front）或最小（back）值）。
在初始化设置了 zIndex 参数时，position 参数的默认值失效，直接使用该值作为图层zIndex值。
如果之后手动调用 bringToFront 或 bringToBack 方法，之前手动设置的zIndex失效，自动分配图层 zIndex 值。
-----------------------------渣翻译分割线--------------------------
if set position option, it will auto set the image dom zIndex. (go through all other layers of the same pane, set zIndex to max + 1 (front) or min - 1 (back))
if set zIndex option, it will use the value as the image dom zIndex.
if call bringToFront / bringToBack function, though set zIndex option in initialize, it will auto set the image dom zIndex.

### DIFF
--- a/src/Layers/RasterLayer.js
+++ b/src/Layers/RasterLayer.js
@@ -228,7 +228,7 @@ export var RasterLayer = Layer.extend({
 
             if (this.options.position === 'front') {
               this.bringToFront();
-            } else {
+            } else if (this.options.position === 'back') {
               this.bringToBack();
             }
             
--- a/src/Layers/RasterLayer.js
+++ b/src/Layers/RasterLayer.js
@@ -30,7 +30,11 @@ export var RasterLayer = Layer.extend({
   onAdd: function (map) {
     // include 'Powered by Esri' in map attribution
     setEsriAttribution(map);
-
+    
+    if (this.options.zIndex) {
+      this.options.position = null;
+    }
+    
     this._update = Util.throttle(this._update, this.options.updateInterval, this);
 
     map.on('moveend', this._update, this);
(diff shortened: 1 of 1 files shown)


jgravois:
thank you for this @appleshowc! but my gut reaction is that it is preferable to use the [map pane](http://leafletjs.com/reference-1.3.0.html#map-pane) functionality that is present in Leaflet itself _instead_ of relying on `z-index` to control layer order.

appleshowc:
Thanks for replay!
I have some layers managed by group, contains tile, dynamic and other type layer in one group. Groups order can be adjusted, but layers order in one group are fixed.
If one layer in one map pane, there are too many panes to manage, and it's hard to control the group order.
For convenience of adjusting the order between groups or layers in one group, I choose one group in one map pane, and use z-index to fix layers order in one group.
It's a suggestion~

appleshowc:
I'm glad to receive your reply @jgravois !  
And of course yes!
**BUT** there is still [errors](https://travis-ci.org/Esri/esri-leaflet/builds/401699044?utm_source=github_status&utm_medium=notification) that I don't know how to fix. 
It confused me. What I submitted is just "RasterLayer.js". However, the errors occurs in other places...sad...

jgravois:
> the errors occur in other places...

we can blame #1038 for that. there is nothing that would make me happier than learning how to fix _that_ particular problem.

i restarted the build (twice) and i'm seeing green now.

jgravois:
it took me _way_ longer than it should have to pull down this code and take a close look at it.

thanks for your patience! its a great patch.

Issue #1090, opened 2018-04-17
token repeating in url


jeremyzc:
I'm using the sample https://esri.github.io/esri-leaflet/examples/switching-basemaps.html as a base for a project that I'm working on.

I've modified the sample to pass in a token;

```
 layer = L.esri.basemapLayer(basemap, {
      token: 'abc'
    });
```

However if I select the same basemap type twice I notice that the token gets repeated in the URL. From the chrome developer tools - token=abc appears twice.

![image](https://user-images.githubusercontent.com/17418983/38901953-675e3ae0-42f2-11e8-8efd-6ad80c9992f2.png)

I think this is happening because when the basemap layer is initialised it appends the token to the urlTemplate which is static if one of the standard map types are used - from the esri-leaflet initialize function;

```
    if (this.options.token) {
      config.urlTemplate += ('?token=' + this.options.token);
    }
```

jgravois:
interesting catch!

do you have any interest in submitting a PR to make that check a little more intelligent?

also, why append a token to `basemapLayer`. are you just doing it to track requests?

jeremyzc:
Reading https://github.com//Esri/esri-leaflet/wiki/FAQ#why-use-esri-leaflet under _Which services require authentication_ it states;

> Authentication is also required if you want to use ArcGIS Online services that have not been shared publicly

It also notes that basemaps do not require authentication so it does seem ambiguous for a private / commercial app.

I think I'm doing the right thing by getting a token from ArcGis (App login) and then passing it in. basemaps are the only service being used in this app at present. Although I'm assuming that the token does not get validated or tracked.

If tokens do not need to be passed in then I guess there is no issue.

I can have a go at a PR, I'm not that familiar with the code so it may take me a while. 
First thought was to not replace the token if one was already present in the urlTemplate string however that won't work if the token changes.
Second thought was manipulate urlTemplate to remove and replace the token.
Third thought was to keep a separate "urlBase" string and then compose a new urlTemplate from the urlBase and token on initialization.

jgravois:
basemap requests don't require authentication. passing a token is purely a mechanism to circumvent rate-limits that are applied to anonymous requests.

https://github.com/Esri/esri-leaflet/pull/800#issue-77352391

> First thought was to not replace the token if one was already present in the urlTemplate string however that won't work if the token changes.

that'd be sufficient to me. we can introduce more complex logic later if anyone ever actually gets bit by a stale token being passed through.

PR #1092, opened 2018-04-24
Prevent token repeating in url

octavm:
Something like this? Fix for #1090


octavm:
Ok, thanks for the advice. What about now?

PR #1099, opened 2018-06-21
Custom request headers & onMetadata callback

delatal:


src/Request.js
```diff
@@ -92,13 +102,15 @@ function xmlHttpPost (url, params, callback, context) {
...
+
+  addHeaders(httpRequest, headers);
   httpRequest.setRequestHeader('Content-Type', 'application/x-www-form-urlencoded; charset=UTF-8');
   httpRequest.send(serialize(params));
 
   return httpRequest;
 }
 
-function xmlHttpGet (url, params, callback, context) {
+function xmlHttpGet (url, params, callback, context, headers) {
```
jgravois: thanks for your contribution @delatal!

i assume you tacked on `headers` as an additional method argument to ensure backwards compatibility, but my first instinct is that its placement in the signature is awkward (given that it is _preceded_ by an optional context binding).

what header/s do you need to pass along in your own requests exactly?

delatal: I just need to add an Authorization header.

jgravois:
i appreciate this contribution but my gut reaction is that the change shouldn't be merged as-is and i don't really have any good suggestions that wouldn't be _breaking_. 

either way, its a good example for others to follow if they find themselves in a similar situation.

delatal:
OK, I understand.

PR #1100, opened 2018-06-21
add support for Firefly Imagery Basemap

pmacMaps:
My first stab at this kind of thing. Couldn't figure out how to test to make sure it works.  So, don't be surprised if there are issues. 

Trying to get my feet wet with contributing actual code instead of bug reports.

### DIFF
--- a/spec/Layers/BasemapLayerSpec.js
+++ b/spec/Layers/BasemapLayerSpec.js
@@ -57,7 +57,7 @@ describe('L.esri.BasemapLayer', function () {
   });
 
   it('can return valid basemaps', function () {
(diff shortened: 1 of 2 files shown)


jgravois:
very cool @pmacMaps!

> Couldn't figure out how to test to make sure it works.

after following the instructions [here](https://github.com/Esri/esri-leaflet#development-instructions) and making changes to the sourcecode, running [`npm start`](https://github.com/Esri/esri-leaflet/blob/c6f4dec79be4d819932ca783b92f9c41d5098f4c/package.json#L89) will spin up a local webserver that automatically recompiles the plugin every time another edit is detected.

this means that you can make a copy of debug/sample.html to quickly test things out.

does that make sense?

pmacMaps:
Yep, Will keep in mind for the future.

jgravois:
thanks for this! i made one small tweak to ensure that the stock attribution is updated when users pan/zoom.

screenshot 2018-06-29 16 38 15

PR #1101, opened 2018-06-22
add world physical as a base map option

pmacMaps:
Saw this ArcGIS Online basemap was still left hanging.

### DIFF
--- a/src/Layers/BasemapLayer.js
+++ b/src/Layers/BasemapLayer.js
@@ -183,6 +183,25 @@ export var BasemapLayer = TileLayer.extend({
           maxZoom: 19,
           attribution: 'Esri, DigitalGlobe, GeoEye, Earthstar Geographics, CNES/Airbus DS, USDA, USGS, AeroGRID, IGN, and the GIS User Community'
         }
+      },
+      WorldPhysical: {
+        urlTemplate: tileProtocol + '//{s}.arcgisonline.com/arcgis/rest/services/World_Physical_Map/MapServer/tile/{z}/{y}/{x}',
+        options: {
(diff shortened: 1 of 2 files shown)


jgravois:
thx @pmacMaps! I'll be out of the office until late next week but I'll review both your pull requests when I get back.

src/Layers/BasemapLayer.js
```diff
@@ -183,6 +183,25 @@ export var BasemapLayer = TileLayer.extend({
           maxZoom: 19,
           attribution: 'Esri, DigitalGlobe, GeoEye, Earthstar Geographics, CNES/Airbus DS, USDA, USGS, AeroGRID, IGN, and the GIS User Community'
         }
+      },
+      WorldPhysical: {
```
jgravois: for consistency with the others, just use `Physical`

src/Layers/BasemapLayer.js
```diff
@@ -183,6 +183,25 @@ export var BasemapLayer = TileLayer.extend({
...
+        urlTemplate: tileProtocol + '//{s}.arcgisonline.com/arcgis/rest/services/World_Physical_Map/MapServer/tile/{z}/{y}/{x}',
+        options: {
+          minZoom: 1,
+          maxZoom: 8,
+          subdomains: ['server', 'services'],
+          attribution: 'U.S. National Park Service'
+        }
+      },
+      WorldPhysicalLabels: {
```
jgravois: i'd skip adding `PhysicalLabels`. folks can just use `TerrainLabels` (which points at the same service).

jgravois:
very cool. thank you!

PR #1122, opened 2018-08-17
Add support for proxy option on BasemapLayer

dangowans:
To help support loading ESRI layers behind Intranet-only restrictions.


dangowans:
Wrote a test in #1123 doing a check similar to the check for "token", but I'm not sure it will work well as the proxy script won't exist.

jgravois:
(a belated) thanks for the contribution @dangowans!

PR #1126, opened 2018-10-05
Dependencies update

gavinr:
Updated misc dev dependencies and also removed `isparta` package


jgravois:
thanks for this @gavinr! i've restarted the Travis job a few times and only the ImageMapLayer tests are running consistently ☹️.

my first instinct is to update karma.conf.js to use a glob and load *all* the tests into memory and see what that does.

```js
files: [
  'node_modules/leaflet/dist/leaflet.css',
  'node_modules/leaflet/dist/leaflet-src.js',
  'dist/esri-leaflet-debug.js',
  'spec/**/*Spec.js'
],
```
https://github.com/Esri/esri-leaflet/blob/master/karma.conf.js#L16-L31

gavinr:
ok, i will make that change and we can try it out

PR #1147, opened 2019-03-28
Fire loading event before export request

### LINKED ISSUE #1146: Missing event before export request in dynamicMapLayer
Now when using dynamicMapLayer, 'loading' event fires only after export request which sometimes takes quite long. So it is not possible to show a loader before export request happens.
It would be very useful to fire 'loading' event (or introduce another event) before export request.

ogix:
Applies both to DynamicMapLayer and ImageMapLayer.
Fixes #1146

### DIFF
--- a/src/Layers/RasterLayer.js
+++ b/src/Layers/RasterLayer.js
@@ -296,7 +296,7 @@ export var RasterLayer = Layer.extend({
 
     if (params) {
       this._requestExport(params, bounds);
-      
+
       this.fire('loading', {
         bounds: bounds
       });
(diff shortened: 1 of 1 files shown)


jgravois:
Thanks @ogix! It looks like the failed build is just a lint error.

> /home/travis/build/Esri/esri-leaflet/src/Layers/RasterLayer.js
  299:1  error  Trailing spaces not allowed

ogix:
Removed the spaces.

jgravois:
awesome. thank you for your contribution!

PR #1152, opened 2019-05-24
added warning if setTimeRange() is called without a `timeField`

gavinr:
added warning per @jgravois comment https://github.com/Esri/esri-leaflet/issues/1148#issuecomment-479708440. I'm not quite sure if `return [];` is the correct thing to do [here](https://github.com/Esri/esri-leaflet/pull/1152/files#diff-b59294fc1c59924a39c0eb1607d4cc9cR368) ... open to suggestions.

PR #1157, opened 2019-07-11
Test time range on features before adding to map

pjbiogit:
Test if feature is within time range before any addLayer call.

### DIFF
--- a/src/Layers/FeatureLayer/FeatureLayer.js
+++ b/src/Layers/FeatureLayer/FeatureLayer.js
@@ -89,7 +89,7 @@ export var FeatureLayer = FeatureManager.extend({
       var layer = this._layers[geojson.id];
       var newLayer;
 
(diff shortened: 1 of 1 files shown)


jgravois:
@gavinr whenever you find time to review this, make sure to explore its relationship with https://github.com/Esri/esri-leaflet/pull/1152 

@pjbiogit, what `timeFilterMode` was causing incorrect behavior? for the default (`server`), the client code should be able to safely skip filtering when making new requests.

pjbiogit:
I attempted using both server and client for the `timeFilterMode` and got the same results prior to my changes. Following changes, both modes function the same way.

patrickarlt:
@gavinr this looks pretty good to me. I say we should merge this and cut a new release.

gavinr:
thanks @pjbiogit!

PR #1159, opened 2019-07-12
swap in 'esri-oss' as the recommended stack overflow tag

jgravois:
When we started working on [ArcGIS REST JS](https://github.com/Esri/arcgis-rest-js), I seeded a generic <kbd>esri-oss</kbd> Stack Overflow / GIS Stack Exchange tag to make it easier for maintainers to keep an eye on questions from folks asking for help using our different open source projects on GitHub.

Given the relatively modest volume of questions posted and tagged with <kbd>esri-leaflet</kbd>, I'd recommend just combining the two.

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,7 +132,7 @@ http://esri.github.io/esri-leaflet/plugins/
 
 If something isn't working the way you expected, please take a look at [previously logged issues](https://github.com/Esri/esri-leaflet/issues?labels=FAQ&milestone=&page=1&state=closed) that resolve common problems first.  Have you found a new bug?  Want to request a new feature?  We'd love to hear from you.  Please let us know by submitting an [issue](https://github.com/Esri/esri-leaflet/issues).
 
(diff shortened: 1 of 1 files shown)


gavinr:
@jgravois over on arcgis-rest-js, we're [linking to the GIS stackexchange](https://github.com/Esri/arcgis-rest-js/blame/master/README.md#L95), but in this PR you're linking to stackoverflow. Should we change one of them to be consistent?

Note that there [appears to be no current questions with that tag on GIS Stackexchange](https://gis.stackexchange.com/questions/tagged/esri-oss), but [stackoverflow has 18 questions](https://stackoverflow.com/questions/tagged/esri-oss).

jgravois:
> Should we change one of them to be consistent?

I'd definitely be okay with that.

patrickarlt:
Sounds good to me. I'll open a PR against REST JS.

PR #1170, opened 2019-10-10
Release notes

gavinr:
Adding some notes about the release process to make it smoother going forward.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,14 @@
+- Check out the repository and pull the latest updates on MASTER (origin or your fork should work)
+- Update the package version in `package.json`.
+- Update `CHANGELOG.md` - remember to update the links at the bottom of the document.
+- Run `npm install`
+- Commit with message `:package: X.X.X` (replacing with version)
+- Push commit to master
+- Run `scripts/release.sh`
+
+
+After that's all complete, you can do these later at your leisure:
+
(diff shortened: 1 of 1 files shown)


RELEASE.md
```diff
@@ -0,0 +1,15 @@
+- Check out the repository and pull the latest updates on MASTER (origin or your fork should work)
+- Update the package version in `package.json`.
+- Update `CHANGELOG.md` - remember to update the links at the bottom of the document.
+- Run `npm install`
```
jgravois: did you mean `npm version`?

RELEASE.md
```diff
@@ -0,0 +1,15 @@
...
+- Commit with message `:package: X.X.X` (replacing with version)
+- Push commit to master
+- Run `npm whoami`, making sure you're logged in properly to NPM. If not, `npm login`.
+- Run `scripts/release.sh`
```
jgravois: nothing wrong with what you have, but `npm run release` is handy too.

gavinr: Good point. Updated. Thanks @jgravois

RELEASE.md
jgravois: > One time I did `npm version` and then later `gh-release` failed

now that I think about it, that actually makes sense. `gh-release` helps with creating a tag (and release) on GitHub using the information in the CHANGELOG. it expects that you will have bumped the version number in your package.json file manually, but _not_ already tagged a version.

`npm version` bumps your package.json and creates a (local) tag.

Issue #1173, opened 2019-11-07
_featureWithinTimeRange() missing a case


pjbiogit:
* Browser and version:

All

* Version of Leaflet (`L.version`):

1.5.1

* Version of esri Leaflet (`L.esri.VERSION`):

2.3.1

I am using the time filtering ability of feature layers to filter specific layers by a start and end time on each field. For this, I have defined the `timeField` with both a `start` and an `end` and defined the layer as having a `to` and a `from` to be used as the time range over which a feature should appear. 
The function `featureWithinTimeRange(feature)` , which I have copied below, is used to determine if a feature falls within a time range. When a feature has a start and end time it evaluates to true if either the start or end time of the feature falls within the layer's set time range. I believe that this is operating under the assumption that there are 5 cases:
* A feature's `start` and `end` fall before the `from` time.
* A feature's `start` falls before the `from` time and its `end` falls between `from` and `to`.
* A feature's `start` and `end` times fall between the `from` and `to` times. 
* A feature's `start` time falls between the `from` and `to` times, and its `end` time falls after `to`.
* A feature's `start` and `end` fall after the `to` time.

The middle three cases should evaluate to true, and are implemented in the code below:
```
_featureWithinTimeRange: function (feature) {
	    if (!this.options.from || !this.options.to) {
	      return true;
	    }

	    var from = +this.options.from.valueOf();
	    var to = +this.options.to.valueOf();

	    if (typeof this.options.timeField === 'string') {
	      var date = +feature.properties[this.options.timeField];
	      return (date >= from) && (date <= to);
	    }

	    if (this.options.timeField.start && this.options.timeField.end) {
	      var startDate = +feature.properties[this.options.timeField.start];
	      var endDate = +feature.properties[this.options.timeField.end];
	      return ((startDate >= from) && (startDate <= to)) || ((endDate >= from) && (endDate <= to));
	    }
	  },
```
However, I believe that there should be a sixth case:
* The `from` and `to` times fall between the feature's `start` and `end` times.

In the current code, this will not evaluate to true, however in my application I would like the feature to persist over its entire time range even if the `from` and `to` represent a time within that range. 
In this case, the final return statement would have one more or case as follows:
`return ((startDate >= from) && (startDate <= to)) || ((endDate >= from) && (endDate <= to)) || ((startDate <= from) && (endDate >= to));`
Please let me know if there was a reason to exclude this case and I can make the necessary changes to my own code to get the proper functionality that I'm looking for. 
[...]

PR #1174, opened 2019-11-07
Add case to _featureWithinTimeRange(feature)

pjbiogit:
Add a case for when the feature's time range contains the time range given in the to and from options for the layer. This way, a feature will always show up during its specified time range.

#1173

### DIFF
--- a/spec/Layers/FeatureLayer/FeatureManagerSpec.js
+++ b/spec/Layers/FeatureLayer/FeatureManagerSpec.js
@@ -661,6 +661,22 @@ describe('L.esri.FeatureManager', function () {
       },
       'id': 2
     })).to.equal(false);
+    
+    expect(layer._featureWithinTimeRange({
+      'type': 'Feature',
+      'geometry': {
+        'type': 'Point',
+        'coordinates': [-122.673345, 45.537188]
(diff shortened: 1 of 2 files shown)


gavinr:
@pjbiogit Thank you very much for the issue logged and PR!

We would like to include a unit test case update that covers your case illustrated here (I _think_ it would go in [here](https://github.com/Esri/esri-leaflet/blob/master/spec/Layers/FeatureLayer/FeatureManagerSpec.js)) - are you able to find some time to do that (feel free to just add an additional commit onto this branch/PR)? If not, one of us will try to get to it.

pjbiogit:
I believe this is all I needed to add, but let me know if you need anything else. Added a case that should evaluate to true when the start time is set to before the from time and the end time is set to after the to time.

pjbiogit:
Attempted to add the test case but it seems something has gone wrong. I'm not very familiar with this so let me know if you see where I've gone wrong.

jgravois:
no worries. its just a lint error: https://travis-ci.org/Esri/esri-leaflet/builds/609303742

gavinr:
thanks @pjbiogit!

Issue #1180, opened 2020-01-15
Dynamic layer cannot use both f:image and token


jptheripper:
* Browser and version:

Chrome 51

* Version of Leaflet (`L.version`):

0.7.7, 1.0.0?

* Version of esri Leaflet (`L.esri.VERSION`):

1.0.4, 2.0.3, `master`?

Steps to reproduce the error:

1. create map
2. add server authenticated service that requires password. 
3. add parameter f:'json'
4. add parameter token: response.token

What happens is the image parameter is ignored and request is sent with type f=json

I was expecting f=image to be respected.  Feature service only loads half the time due to internal load balancer not having request on /arcgisoutput (which is why we need to use the f=image parameter)

https://jsbin.com/hiqequg/edit?html,output

jwasilgeo:
Some more background and historical info would be helpful here from previous contributors to [`./src/Layers/DynamicMapLayer.js`](https://github.com/Esri/esri-leaflet/blob/master/src/Layers/DynamicMapLayer.js).

It appears to me that this is acting as designed in:

- https://github.com/Esri/esri-leaflet/blob/6cf758751e8c34b61415254fa23ec109400e8ec3/src/Layers/DynamicMapLayer.js#L23-L25
- https://github.com/Esri/esri-leaflet/blob/6cf758751e8c34b61415254fa23ec109400e8ec3/src/Layers/DynamicMapLayer.js#L177

jgravois:
the`else` block below needs a `token` check similar to the `if` block above it, except it needs to optionally append the parameter to the original url.

https://github.com/Esri/esri-leaflet/blob/6cf758751e8c34b61415254fa23ec109400e8ec3/src/Layers/DynamicMapLayer.js#L194-L197

besides the service in @jptheripper's jsbin, you can also test with the one below.

username: user1
password: user1

http://sampleserver6.arcgisonline.com/arcgis/rest/services/Wildfire_secure_ac2/MapServer/export?token=foo&...

jptheripper:
This is absolutely perfect.  We can confirm once this functionality is applied that f='image' and token work together as desired.

jptheripper:
Sorry, I am new to the whole process.  How do we get the change applied to the source code?

jgravois:
> How do we get the change applied to the source code?

[pull requests](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/about-pull-requests) are _always_ welcome.

[...]

Issue #1153, opened 2019-06-05
DynamicMapLayer and Proxy JSON returned results in Proxy Request ?undefined


larotech:
* Browser and version:
Chrome Version 71.0.3578.98

* Version of Leaflet (`L.version`):
1.5.1

* Version of esri Leaflet (`L.esri.VERSION`):
v2.2.4 - Wed Mar 20 2019 16:03:22 GMT-0700 (Pacific Daylight Time)

* ArcGIS Version
10.31

Steps to reproduce the error:

We have two ArcGIS Server environments, one for DEV and one for PROD. 
The DEV MapService reports: Supports Dynamic Layers: false
The PROD MapService reports: Supports Dynamic Layers: true

The MapService Layer '19' is being shown using:
```
L.esri.dynamicMapLayer({
                    url: overlayUrl,
                    proxy: "/proxy.ashx",
                    opacity: 1,
                    useCors: true,
                    layers: [19],
                    transparent: true,
                    format: 'png32'
                }).addTo(map);
```

When the overlayUrl is pointing to the MapService in DEV which has "Supports Dynamic Layers: false" the JSON output contains a parameter called "href" which contains the url to the export image. 

When the overlayUrl is pointing to the MapService in PROD which has "Supports Dynamic Layers: true" the JSON output contains a parameter called "imageData" which is a base64 encode of the export image. 

The ESRI Leaflet JavaScript code at: https://github.com/Esri/esri-leaflet/blob/adedb89e965fdbdb08c80c771bf245fca90f69c9/src/Layers/DynamicMapLayer.js#L185 assumes there is an output property in the returned JSON called "href" which in the case of my MapService in PROD, its not returning an "href" but instead an "imageData" property containing a base64 image. What happens in this case is this._renderImage is passed a value of undefined which results in a request of /proxy.ashx?undefined 

I propose a check for response.href or something similar to get around this issue:
```
if (this.options.proxy && response.href) {
          response.href = this.options.proxy + '?' + response.href;
        }
        if (response.href) {
          this._renderImage(response.href, bounds);
        } else {
          this._renderImage(response.imageData, bounds, response.contentType);
        }
```
This would then allow the renderImage method to look at response.imageData as href == undefined. This issue only happens for resources that need to be passed thru a proxy.

Regards,
Chris

jgravois:
related: https://github.com/Esri/esri-leaflet/issues/847

patrickarlt:
[...]

PR #1181, opened 2020-01-17
1180-image-and-token

jwasilgeo:
1. This was initially related to #1180 to allow **both** an optional `token` and `f: image` for a `DynamicMapLayer` instance.  It seems that the conditional [override](https://github.com/Esri/esri-leaflet/compare/1180-image-and-token?expand=1#diff-a7be8ad7c53337df7dc83d076f320957L23-L25) of `f` in the `initialize` method was not allowing both of these options to persist.  But, I still [don't know what the original intent of this was](https://github.com/Esri/esri-leaflet/issues/1180#issuecomment-574885562) and it would be easier to rest at night if we could find out.  The similarly-structured `ImageMapLayer` does not do this in its `initialize`.

2. Please also reference changes introduced a few months ago for issue #1153 that involved related blocks of code in `DynamicMapLayer`'s `_requestExport`.  I'm leaving this as a breadcrumb here in case we find in the future that we need to add the same extra checks in `ImageMapLayer`'s `_requestExport`.

3. Along the way I standardized similarities I noticed between `DynamicMapLayer` and `ImageMapLayer` in their usage of `this.options` vs `params` in their `_buildExportParams` and `_requestExport` methods.

4. Finally, I added an optional _prepending_ of `params.proxy` when `f` is `image` for both layer types.  
I do not have a way to easily test this functionality.  We may want to strike it from this PR and reconsider it in the future if it is reported as an issue that someone can't use both `f: image` and a `proxy`.  Note that the _appending_ of `params.token` was already successfully assumed to occur in the code.
    - https://github.com/Esri/esri-leaflet/compare/1180-image-and-token?expand=1#diff-a7be8ad7c53337df7dc83d076f320957R201-R204
    - https://github.com/Esri/esri-leaflet/compare/1180-image-and-token?expand=1#diff-92a727b8d8cc44ec622b0a0335471d15R207-R210

I can consider updating related tests once we have a chance to talk a bit about these changes.

I also added 3 new debug HTML pages to verify my work so far.

cc @jgravois @jptheripper

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,6 +34,7 @@ debug/*.json
 debug/archive/
 !debug/sample.html
 !debug/sample-dynamic-map-layer.html
+!debug/sample-dynamic-map-layer-server-auth.html
 !debug/sample-image-map-layer.html
 
 # Docs Build
--- a/.gitignore
+++ b/.gitignore
@@ -33,6 +33,8 @@ debug/*.js
 debug/*.json
 debug/archive/
 !debug/sample.html
+!debug/sample-dynamic-map-layer.html
+!debug/sample-image-map-layer.html
 
 # Docs Build
 site/build
--- a/debug/sample-dynamic-map-layer-server-auth.html
+++ b/debug/sample-dynamic-map-layer-server-auth.html
@@ -1,0 +1,103 @@
+<!DOCTYPE html>
+<html>
+
+<head>
+  <meta charset=utf-8 />
+  <title>Esri Leaflet Debugging Sample</title>
+  <meta name='viewport' content='initial-scale=1,maximum-scale=1,user-scalable=no' />
+
+  <!-- Load Leaflet -->
+  <link rel="stylesheet" href="../node_modules/leaflet/dist/leaflet.css" />
+  <script src="../node_modules/leaflet/dist/leaflet-src.js"></script>
+
+  <!-- Load Esri Leaflet from source-->
+  <script src="../dist/esri-leaflet-debug.js"></script>
+
+  <style>
+    body {
+      margin: 0;
+      padding: 0;
+    }
+
+    #map {
+      position: absolute;
+      top: 0;
+      bottom: 0;
+      right: 0;
+      left: 0;
+    }
+
+    #info-pane {
+      position: absolute;
+      top: 10px;
+      right: 10px;
+      z-index: 1000;
+      padding: 1em;
+      background: white;
+    }
+  </style>
+</head>
+
+<body>
+
+  <div id="map"></div>
+  <div id="info-pane" class="leaflet-bar">
+    <label>
+      sample application for debugging a <pre>L.esri.dynamicMapLayer</pre>
+    </label>
+  </div>
+
+  <script>
+    /*
+    make a copy of this file in the same folder if you'd like git to ignore your local changes
+    */
+    var map = L.map('map').setView([40, -100], 4);
+    L.esri.basemapLayer('Topographic').addTo(map);
+
+    function serverAuth(callback) {
+      L.esri.post('https://sampleserver6.arcgisonline.com/arcgis/tokens/generateToken', {
+        username: 'user1',
+        password: 'user1',
+        f: 'json',
+        expiration: 86400,
+        client: 'referer',
+        referer: window.location.origin
+      }, callback);
+    }
+
+    serverAuth(function (error, response) {
+      if (error) {
+        return;
+      }
+
+      var dl = L.esri.dynamicMapLayer({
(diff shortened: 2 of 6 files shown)


jgravois:
> I do not have a way to easily test this functionality.
I can consider updating related tests once we have a chance to talk a bit about these changes.

I'd say these changes are a great candidate for some [TDD](https://en.wikipedia.org/wiki/Test-driven_development).

what you're proposing is adding support for manipulating DynamicMapLayer `f:"image"` requests with:
* a proxy
* a token

right now Travis is reporting a failing test. if you can fix that test and add two more then we'd be (reasonably) sure that the tweaks accomplish what you advertise and won't break any existing use cases.

[this test](https://github.com/Esri/esri-leaflet/blob/c3130441f949dc1125a18b047eea9620c9ed1e5b/spec/Layers/DynamicMapLayerSpec.js) would be a good one to copy/paste and modify. sound good?

jptheripper:
Just to be clear, we are NOT pulling a /ImageServer as type image
we are pulling a dynamic /MapServer as f=image (as opposed to json) to prevent accessing the /arcgisoutput directory

jwasilgeo:
@jptheripper, yes thanks for the clarification. While working on this I wanted to make the same/similar changes to the related `ImageMapLayer.js` for /ImageServer because they are quite similar in nature.

jptheripper:
Perfect thank you. I really appreciate this.

jwasilgeo:
@jgravois I tinkered with and tried to work on the existing test for a but, but in the end I found that can get the test in question passing again by reversing this change:
https://github.com/Esri/esri-leaflet/pull/1181/files#diff-a7be8ad7c53337df7dc83d076f320957L182-R181

Which doesn't quite make sense to me since `token` is available in both `this.options` and `params` at that point in time in the method `_requestExport` _in the case of f: "json"_.  It is added on to `params` just before in the method `_buildExportParams` (and to add to the uncertainty it came from `this.service.options` rather than from `this.options` (which have very similar properties)).

Thus (my bullet 3 above) I had hoped that by replacing `this.options` with `params` in `_requestExport` it would make reading through and understanding that method easier in the future in `DynamicMapLayer` and `ImageMapLayer`. Perhaps in the interest of time it may be simplest to reverse those changes?

---

> what you're proposing is adding support for manipulating DynamicMapLayer f:"image" requests with:
> - a proxy
> - a token

That's correct (for my bullet 4 above), but now I'm thinking about dropping proxy support for f:image since this PR already introduces enough changes and I don't want to hold up #1180 too long unless someone else has more free time for [writing new proxy tests](https://github.com/Esri/esri-leaflet/blob/c3130441f949dc1125a18b047eea9620c9ed1e5b/spec/Layers/DynamicMapLayerSpec.js#L369) (which I found commented out).

However, regarding adding token support for f: "image":
> Note that the _appending_ of `params.token` was already successfully assumed to occur in the code.

That's because `_requestExport` _in the case of f: "image"_ was [only using `params` (no `this.options`)](https://github.com/Esri/esri-leaflet/pull/1181/files#diff-a7be8ad7c53337df7dc83d076f320957L196) and appeared to be doing it fine with the `token` available in the `params` object.

jwasilgeo:
The more I ponder this, the more I see this PR is trying to do too much.  Let's start with @gavinr's new unit tests in PR #1182 and go from there.  Perhaps then we can ultimately focus on only what is needed to satisfy issue #1180.  Again, that's described above as:

> 1. This was initially related to #1180 to allow **both** an optional `token` and `f: image` for a `DynamicMapLayer` instance.  It seems that the conditional [override](https://github.com/Esri/esri-leaflet/compare/1180-image-and-token?expand=1#diff-a7be8ad7c53337df7dc83d076f320957L23-L25) of `f` in the `initialize` method was not allowing both of these options to persist.  But, I still [don't know what the original intent of this was](https://github.com/Esri/esri-leaflet/issues/1180#issuecomment-574885562) and it would be easier to rest at night if we could find out.  The similarly-structured `ImageMapLayer` does not do this in its `initialize`.

My proposed change was to delete those override lines in `initialize`, but an alternative code change to consider could be @jptheripper's [comment here](https://github.com/Esri/esri-leaflet/issues/1180#issuecomment-575185127).

jgravois:
> this PR is trying to do too much.

those are wise words. it will always be possible to change the behavior for proxies in a follow-up pass and limiting a pull request to tackling one thing at a time doesn't just make it easier to write, it makes it easier to read too.

jwasilgeo:
Closing, because #1180 was fixed in PR #1182. We can refer to some of the info in this comment thread later when dealing with proxy support with `f: 'image'` in `DynamicMapLayer` and `ImageMapLayer`.

PR #1182, opened 2020-01-22
Dynamic layer using both f:image and token

gavinr:
This contains a few additional test cases, one of which is failing, illustrating the issue in #1180 ....  I think  :)

@jptheripper @jwasilgeo @jgravois do you think this covers the issue in #1180?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased][unreleased]
 
+### Fixed
+
+* Ensure `DynamicMapLayer` can use both a token and `f: 'image'`. ([#1180](https://github.com/Esri/esri-leaflet/issues/1180))
+
 ## [2.3.2] - 2019-11-13
 
 ### Fixed
--- a/spec/Layers/DynamicMapLayerSpec.js
+++ b/spec/Layers/DynamicMapLayerSpec.js
@@ -366,6 +366,29 @@ describe('L.esri.DynamicMapLayer', function () {
     expect(spy.getCall(0).args[0]).to.match(new RegExp(/http:\/\/services.arcgis.com\/mock\/arcgis\/rest\/services\/MockMapService\/MapServer\/export\?bbox=-?\d+\.\d+%2C-?\d+\.\d+%2C-?\d+\.\d+%2C-?\d+\.\d+&size=500%2C500&dpi=96&format=png24&transparent=true&bboxSR=3857&imageSR=3857&f=image/));
   });
 
+  it('should be able to request an image directly from the export service, with auth', function () {
+    layer = L.esri.dynamicMapLayer({
+      url: url,
+      f: 'image'
+    });
+    var spy = sinon.spy(layer, '_renderImage');
+    layer.authenticate('foo');
+    layer.addTo(map);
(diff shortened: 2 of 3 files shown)


jgravois:
first you write the tests...
then you fix the tests...
and when you fix the tests, you squash the 🐛.

plenty of ways to skin a cat, but one would be for @jwasilgeo to PR against _this_ branch and then merge this one to `master` once Travis is happy.

jwasilgeo:
Thanks for starting up this testing effort, @gavinr.  Much appreciated.  I have pushed 4b84d4c which gets your new tests and all tests to pass.

This is the same change I [described here](https://github.com/Esri/esri-leaflet/pull/1181#issuecomment-577246604) for `DynamicMapLayer.js`.

>... proposed change was to delete those override lines in `initialize` ...

jgravois:
> I have pushed 4b84d4c which gets your new tests and all tests to pass.

there's just one problem. you guys didn't write a proxy test yet.

```js
it('should be able to pass a request for a direct image through a proxy', function () {
  layer = L.esri.dynamicMapLayer({
    url: url,
    f: 'image',
    proxy: './proxy.ashx'
  });
  var spy = sinon.spy(layer, '_renderImage');
  layer.addTo(map);
  expect(spy.getCall(0).args[0]).to.match(new RegExp(/\.\/proxy.ashx\?http:\/\/services.arcgis.com\/mock\/arcgis\/rest\/services\/MockMapService\/MapServer\/export\?bbox=-?\d+\.\d+%2C-?\d+\.\d+%2C-?\d+\.\d+%2C-?\d+\.\d+&size=500%2C500&dpi=96&format=png24&transparent=true&bboxSR=3857&imageSR=3857&f=image/));
});
```

jwasilgeo:
@jgravois thanks for the updated info in your comment, that helps clarify a few things after temporarily trying out your test snippet.

jwasilgeo:
@jgravois @gavinr please take another look when you can.  The `f` override in `initialize` was trimmed down to now **only override** `f` to be `'json'` if a `proxy` is present in `this.options`.  This means we should be able to alleviate #1180 and the `DynamicMapLayer.js` can now use both `f: 'image'` and a `token`.

I plan to deal with adding a test(s) and adding `DynamicMapLayer.js` support for `proxy` and `f: 'image'` in a separate effort, where I will also try to make sure the same behavior and tests are included with the closely related `ImageMapLayer.js`.

jwasilgeo:
@jgravois @gavinr while looking ahead to the future work I mentioned just above, I _re-remembered_ that `DynamicMapLayer.js` never had passing tests to support `f: 'json'` and a `proxy`, despite that this was the only behavior of the layer if a `proxy` was provided in the options.  My [mega comment over here](https://github.com/Esri/esri-leaflet/pull/1181#issuecomment-576498389) points to [a commented-out test](https://github.com/Esri/esri-leaflet/blob/c3130441f949dc1125a18b047eea9620c9ed1e5b/spec/Layers/DynamicMapLayerSpec.js#L369) that maybe we could get to pass. Unfortunately I'm at a standstill with this one after tinkering with it a bit.  Can I get a 2nd pair of 👀 on this please?

The changes proposed in this PR, for `f: 'image'`, don't have any influence over that branch of the logic in the `_renderImage` method.

jgravois:
> `DynamicMapLayer` never had passing tests to support f: 'json' and a proxy
> The changes proposed in this PR, for f: 'image', don't have any influence over that branch of the logic

that is an improvement that would be better tackled in its own PR. if you're stuck on it, I've found that just commenting out the test and explaining where you're at in a draft PR is a good way to frame the discussion.

jwasilgeo:
That's a good plan to me, @jgravois.  Thanks again for taking the time to talk this through with us.  Otherwise are you both still feeling good about this PR as it is?

PR #1189, opened 2020-03-26
Cacheable feature grid

patrickarlt:
@gavinr @jwasilgeo this PR removes the dependency on `leaflet-virtual-grid` and replaces it with a new `FeatureGrid` that makes a repeatable grid so we can have queries be cached similar to the JS API despite maps having different origins. The code is based on a simplified version of Leaflets [`GridLayer`](https://github.com/Leaflet/Leaflet/blob/master/src/layer/tile/GridLayer.js).

You can test this by:

1. Start the dev server an open the debug app
2. Open Dev Tools > Network and make sure caching is ENABLED
3. Refresh the page until you see the `query` calls until you see 304 responses this means the browser has cached the queries
4. Change the initial lat/lng of the map view and refresh the page. You should still see queries with 304 responses.

@gavinr @jwasilgeo can you validate that this works and that all samples still work. Ideally this should go out without any breaking changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,398 +5,410 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased][unreleased]
 
-* Updated `request` to be imported in `BasemapLayer.js` rather than relying on global `L.esri`. (🙏danieloliveira117🙏 [#1191](https://github.com/Esri/esri-leaflet/issues/1191))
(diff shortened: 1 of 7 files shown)


gavinr:
I am just now noticing the unit tests are failing though

gavinr:
This was released in v2.4.0: https://github.com/Esri/esri-leaflet/releases/tag/v2.4.0

PR #1192, opened 2020-03-31
Add request as import (#1191)

### LINKED ISSUE #1191: L.esri is undefined while zooming in/out
I'm using esri-leaflet as an import (webpack) instead extending the global `L`.

When zooming in or out I'm getting the error:
 `TypeError: L.esri is undefined`

The error is in the following line:

https://github.com/Esri/esri-leaflet/blob/6fb127534445424f34ece5ff9d8caac989bd739f/src/Layers/BasemapLayer.js#L305

It should be able to be fixed simply by importing the `request` function instead of using the global `L.esri`.

danieloliveira117:
Replace unnecessary global L.esri with import.
Fixes #1191.

### DIFF
--- a/src/Layers/BasemapLayer.js
+++ b/src/Layers/BasemapLayer.js
@@ -1,5 +1,6 @@
 import { TileLayer, Util } from 'leaflet';
 import { pointerEvents } from '../Support';
+import { request } from '../Request';
 import {
   setEsriAttribution,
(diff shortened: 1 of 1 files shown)


jwasilgeo:
This is a good find, thanks for providing the update.  I did a find-all for other instances in `./src/` for `L.*` and didn't find anything else.

Issue #1198, opened 2020-05-18
CDN version is broken in IE11


destus90:
## Describe the bug

The hosted CDN version is not compatible with ES5 browsers. There are some ES2015 statements shipped from `@terraformer/arcgis`. This may have happened after merging #1194.

## Expected behavior

Should work in IE11. 

## To Reproduce

https://plnkr.co/edit/I7rNXwkWmITFwupi?open=lib%2Fscript.js

Observe the errors in the console.

gavinr:
@destus90 thank you for the report and the PR. I do see that the latest version does not work in IE11. We will work on testing/verifying your PR. In the mean time you can use a previous version as a workaround: https://unpkg.com/esri-leaflet@2.3.3/dist/esri-leaflet.js

Example: https://plnkr.co/edit/8EGLnIKZ5UmPK3xE?open=lib%2Fscript.js

gavinr:
This was addressed in https://github.com/Esri/esri-leaflet/releases/tag/v2.4.1. Thanks!

PR #1199, opened 2020-05-18
transform bundled code with babel for production profile

destus90:
Closes https://github.com/Esri/esri-leaflet/issues/1198

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -1,0 +1,3 @@
+module.exports = {
+  'presets': ['@babel/preset-env']
+};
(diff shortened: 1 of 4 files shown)


gavinr:
I tested locally, it seems like this change allows the production version to work in IE11. Should we add [this](https://github.com/Esri/esri-leaflet/pull/1199/files#diff-afc79ea641469add1aef50585cce7876R10) to `profiles\debug.js` also to facilitate accurate testing?

Issue #1200, opened 2020-05-18
Featuregrid imports `toLatLngBounds` which leaflet exports as `latLngBounds`.


WyssTobi:
## Describe the bug
In the latest version of esri-leaflet, [FeatureGrid.js](https://github.com/Esri/esri-leaflet/blob/master/src/Layers/FeatureLayer/FeatureGrid.js#L3) imports `toLatLngBounds` from leaflet, which is not exported by leaflet 1.6.0. It is exported as `latLngBounds`. This results in an error in our rollup build. See the following logs:

```
[!] Error: 'toLatLngBounds' is not exported by node_modules/leaflet/dist/leaflet-src.esm.js
https://rollupjs.org/guide/en/#error-name-is-not-exported-by-module
node_modules/esri-leaflet/src/Layers/FeatureLayer/FeatureGrid.js (3:2)
1: import {
2:   LatLngBounds,
3:   toLatLngBounds as latLngBounds,
     ^
4:   Layer,
5:   Browser,
Error: 'toLatLngBounds' is not exported by node_modules/leaflet/dist/leaflet-src.esm.js
```
We temporarly fixed it with an automatic replacement of the codeline after the installation of the npm modules.

## Expected behavior

I expect the FeatureGrid.js to import `latLngBounds` instead of `toLatLngBounds`. You can see the export of `toLatLngBounds` as `latLngBounds` in the [leaflet repository](https://github.com/Leaflet/Leaflet/blob/master/src/geo/index.js#L2).

## Environment Information

- Version of Leaflet (`1.6.0`):
- Version of Esri Leaflet (`2.4.0`):
- Your OS: [macOS]
- Browser and Version (error occurs on build time)

## Additional context
We are working on an angular component library which is bundled by rollup and built by bazel.

gavinr:
@WyssTobi thank you for the report. I do not see this error when running `npm run build` in the esri-leaflet project. Could you please advise on a simplified test case for us to see the error?

jgravois:
@WyssTobi must be doing something in his own rollup config to target leaflet's built ES module (leaflet-src.esm.js) as the source to bundle from instead of the raw ES code we pull from (src/Leaflet.js).

since neither @gavinr or i are familiar with bazel, a sample rollup config would indeed be helpful to proceed.

jwasilgeo:
Closing due to inactivity.

nisarmd:
Started seeing this issue with the following versions -  

- **node.js v16.16.0**
[...]

PR #1194, opened 2020-04-16
use @terraformer/arcgis to generate GeoJSON from older ArcGIS services

jgravois:
over the years we've refactored the code that turns ArcGIS Geometries into GeoJSON _more than once_ in this library.

1. originally we inlined a subset of the code from Terraformer
2. later @patrickarlt created a new standalone package https://www.npmjs.com/package/@esri/arcgis-to-geojson-utils
3. by that point, @Esri had _quite a few_ packages and repos that served a similar purpose. and people started getting confused about when they should use what. 

ref: 
* https://github.com/Esri/terraformer-arcgis-parser/issues/44
* https://github.com/Esri/geojson-utils/issues/16

it always irked me that we were maintaining (and sometimes failing to maintain) so much nearly identical code. so i came up with the brilliant idea to publish **even more**. 😝 

see: https://xkcd.com/927

to make a long story long, `@terraformer/arcgis` is pretty much a glorified renaming of `@esri/arcgis-to-geojson-utils`.
to make a long story longer, it comes paired with another package for converting between WKT and GeoJSON and another package for using spatial predicates on GeoJSON. Esri Leaflet doesn't have any use for these, but they're [big in Japan](https://www.youtube.com/watch?v=DdWZKb659K0).

https://www.npmjs.com/package/@terraformer/wkt
https://www.npmjs.com/package/@terraformer/spatial

why rewrite them? mostly to make the functionality in the second two [tree shakable](https://en.wikipedia.org/wiki/Tree_shaking) and to put everything back under the familiar 'Terraformer' umbrella.

my hope is that this will help us archive all our old [Esri/Terraformer](https://github.com/Esri?q=terraformer&type=&language=) code and all the other alternative GeoJSON parsing code we wrote along the way.

only time will tell though.

PR #1201, opened 2020-05-18
update rollup etc. and install the latest (transpiled) @terraformer/arcgis

jgravois:
thank you @destus90 for pointing out the problem with #1194. the error they saw in IE11 would have been caught out at compilation time if we had been running a newer version of rollup to bundle the library.

this PR is an alternative to #1199 that bumps @terraformer/arcgis to a version that is already transpiled and also updates rollup and its plugins to more modern versions so we don't miss similar errors in the future.

```js
john@zuzu esri-leaflet (master) $ npm run build

> esri-leaflet@2.4.0 prebuild /Users/john/code/ezree/esri-leaflet
> mkdirp dist


> esri-leaflet@2.4.0 build /Users/john/code/ezree/esri-leaflet
> rollup -c profiles/debug.js & rollup -c profiles/production.js


src/EsriLeaflet.js → dist/esri-leaflet-debug.js...

src/EsriLeaflet.js → dist/esri-leaflet.js...
created dist/esri-leaflet-debug.js in 908ms
  262 |    * Apache-2.0 */
  263 |
> 264 |   const edgeIntersectsEdge = (a1, a2, b1, b2) => {
      |  ^ Unexpected token: keyword «const»
  265 |     var uaT = (b2[0] - b1[0]) * (a1[1] - b1[1]) - (b2[1] - b1[1]) * (a1[0] - b1[0]);
  266 |     var ubT = (a2[0] - a1[0]) * (a1[1] - b1[1]) - (a2[1] - a1[1]) * (a1[0] - b1[0]);
  267 |     var uB = (b2[1] - b1[1]) * (a2[0] - a1[0]) - (b2[0] - b1[0]) * (a2[1] - a1[1]);
[!] (plugin uglify) Error: Unexpected token: keyword «const»
```
as you can see from the error, its only `uglify()` that was complaining about the ES2015 syntax. that's why @destus90 targeted the production build specifically.

either way, i feel bad for introducing that bug 🐛. i forgot that i'd sprinkled a little extra ES sugar when i ported arcgis-to-geojson-utils. ~~the good thing about @destus90's fix (and this one) is that they set you up to start using some more modern JS in this library and have it transpiled down for older browsers too if you ever feel like it.~~

i opened up this PR to get a little bit closer to the root of the problem, but also because i thought it would uncover the error to described in #1200.

i was only successful on the first front.

### DIFF
--- a/profiles/base.js
+++ b/profiles/base.js
@@ -1,5 +1,6 @@
-import json from 'rollup-plugin-json';
-import nodeResolve from 'rollup-plugin-node-resolve';
+import json from '@rollup/plugin-json';
+import nodeResolve from '@rollup/plugin-node-resolve';
+import babel from '@rollup/plugin-babel';
 
 var pkg = require('../package.json');
 var copyright = '/* ' + pkg.name + ' - v' + pkg.version + ' - ' + new Date().toString() + '\n' +
@@ -17,7 +18,9 @@ export default {
       browser: false,
       extensions: [ '.js', '.json' ]
     }),
-    json()
+    json(),
+    // use babel to transpile ES code
+    babel({ babelHelpers: 'bundled' })
   ],
 
   output: {
@@ -26,7 +29,8 @@ export default {
     name: 'L.esri',
     globals: {
       'leaflet': 'L',
-      'esri-leaflet': 'L.esri'
+      'esri-leaflet': 'L.esri',
+      '@terraformer/arcgis': 'Terraformer'
     },
     sourcemap: true
   }
--- a/profiles/base.js
+++ b/profiles/base.js
@@ -1,6 +1,5 @@
 import json from '@rollup/plugin-json';
 import nodeResolve from '@rollup/plugin-node-resolve';
-import babel from '@rollup/plugin-babel';
 
 var pkg = require('../package.json');
 var copyright = '/* ' + pkg.name + ' - v' + pkg.version + ' - ' + new Date().toString() + '\n' +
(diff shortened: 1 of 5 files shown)


profiles/base.js
```diff
@@ -26,7 +29,8 @@ export default {
     name: 'L.esri',
     globals: {
       'leaflet': 'L',
-      'esri-leaflet': 'L.esri'
+      'esri-leaflet': 'L.esri',
+      '@terraformer/arcgis': 'Terraformer'
```
jgravois, line 33: this solves an unrelated rollup nag.

```bash
Use output.globals to specify browser global variable names corresponding to external modules
/Users/john/code/ezree/esri-leaflet/node_modules/rollup/dist/rollup.js (guessing 'arcgis')
```

patrickarlt:
Personally my preference would be to not transpile Esri Leaflet at all and resolve this in `@terraformer/arcgis`. `@terraformer/arcgis` should only ship ready-to-use already compiled code anyway so my preference would be to fix the build process there first to ensure Esri Leaflet can get dependencies it can use.

jgravois:
> @terraformer/arcgis should only ship ready-to-use already compiled code anyway

i've been considering it a feature that @terraformer/arcgis ships both a UMD and a flat ESM module that can be tree shaken by consumers like Esri Leaflet.

it sounds like you're saying that the ESM just shouldn't need to be transpiled. i will make it so.

jgravois:
alright, i just pushed another commit to revert the babel step here and 

what's left
* bump @terraformer/arcgis to the latest (complete with internal transpilation)
* bump rollup and friends to latest
* fix rollup nag

sorry this turned into a dogfooding session, but i appreciate you helping me improve terraformer@latest!

i don't have access to a box to test IE11 anymore, so help on that front would be appreciated.

PR #1206, opened 2020-06-19
Feature/fetch everything

francharbo:
I was quite frustrated to see that all my features weren't downloaded using feature layer and cluster (wrong number in clusters)

That's why I made some adjustements you can see here.
I would like to know if it was possible to integrate this ?
maybe I do not grasp the entirety of the problem though


francharbo:
I added some tests and made the routine not affects the rest of the code.
Does someone know how this behaviour works in ArcGIS jsAPI 4.x ? It looks like it's not looping the same way. It could be worth trying to understand it maybe ?

I did some benchmarks using a custom server (around 18000 points with a jointure)
By doing some benchmarks I saw that ArcGIS js 4.x was making something like 89 requests, but everything was downloaded in 7 sec.
However using this PR it takes about 10 sec to load everything.
Does someone have any ideas ?

francharbo:
@patrickarlt @jwasilgeo 
Sorry to come back to you again, but I'd like to know if there is a chance that this request will be integrated in esri-leaflet

francharbo:
@patrickarlt 
I realised that I mixed my two branches and thus there is also the fix for cellEnter (#1205 )
Do you want me to remove this from this PR ?

francharbo:
Thank you for the feedback !

I didn't think about the time needed to transform the geojson to json, nice job !

Another thing I was wondering, is about Travis (not very much about the algorithm itself thus).

In my first commit I tried using default parameter for offset in _buikdQuery : 
```js
 _buildQuery: function (bounds, offset = 0) {
    var query = this.service
      .query()
      .intersects(bounds)
      .where(this.options.where)
      .fields(this.options.fields)
      .precision(this.options.precision);
      .precision(this.options.precision)
      .offset(offset);
```
But this won't pass the Travis test.

Do you prefer keeping the syntax as it is in the PR with the check ?
**in _buildquery :** 
```js
    if (this.options.fetchAllFeatures && !isNaN(parseInt(offset))) {
      query = query.offset(offset);
    }
```
+
**in _requestFeatures:**
```js
_requestFeatures: function (bounds, coords, callback, offset) {
    this._activeRequests++;

    // default param
    offset = offset || 0;
```

patrickarlt:
@francharbo My preference would be to NOT use the ES 2016 default syntax because we don't compile the Esri Leaflet source except for `import`/`export`. So the `offset = offset || 0` is the correct way to do this.

I would merge down the changes I made in https://github.com/Esri/esri-leaflet/compare/Esri:master...fetch-everything-proposal?expand=1 with `git pull origin fetch-everything-proposal` and we can merge this.

src/Tasks/Query.js
```diff
@@ -126,15 +126,15 @@ export var Query = Task.extend({
     this._cleanParams();
 
     // services hosted on ArcGIS Online and ArcGIS Server 10.3.1+ support requesting geojson directly
-    if (this.options.isModern || isArcgisOnline(this.options.url)) {
+    if (this.options.isModern || (isArcgisOnline(this.options.url) && this.options.isModern === undefined)) {
```
gavinr, line 129: @patrickarlt what if someone sets `isModern` to `false`? Should we check that here too or are you intentionally only targeting the case when the user does not set it?

francharbo: If you check also if `isModern === false`, you are not letting the user ask for JSON instead of geoJSON if he wants to ([cf. changelog](https://github.com/Esri/esri-leaflet/blob/master/CHANGELOG.md#added-5))
(That's the reason I found behind this)

gavinr: the `isModern === false` will get caught in the first part of the OR clause, oops - missed that.

francharbo:
@patrickarlt 
I was looking at geojson querying on feature service and I saw that it returned `exceededTransferLimit`.

So my question is : Is it an unexpected behavior or may we be able to use the fetchAll feature with geojson as well ?

I was using a feature service published in a 10.4 server with ArcMap.
image

patrickarlt:
@francharbo that looks like it will work but i want to check with some Esri staff because checking this on https://services.arcgis.com/RmCCgQtiZLDCtblq/arcgis/rest/services/LACPW_Sanitary_Sewer_Network/FeatureServer/1/query?outFields=*&where=1%3D1&f=geojson it looks like this is `response.properties.exceedstransferLimit` in the request I was testing and `request.exceedsTransferLimit` in your request. I suppose you could check if their of those is true and continue making requests as long as one is.

francharbo:
Do you have any answer about this or should I just implement it and remove the need for geojson ?

gavinr:
@francharbo we are still looking into this, sorry for the delay. Should have some more info soon.

patrickarlt:
@francharbo after some checking with the Server and Online teams you will need to check both `response.properties.exceedstransferLimit` and `response.exceedsTransferLimit` if either is true, continue to load more features. The server and online implementations differ.

//cc @gavinr @jwasilgeo

gavinr:
Closing this in favor of cleaned-up branch in #1229

PR #1211, opened 2020-07-08
Fix bugs with feature layer setWhere

### LINKED ISSUE #875: setWhere callback is not called
Changing the where condition on a newly created layer doesn't let the callback to be executed.

```
var layer = L.esri.featureLayer({
    url: "link",
    where: "1=1",
});
layer.addTo(map);
layer.setWhere("1=2", function(){
    console.log("callback called");
});
```

The callback is correctly executed if the setWhere is executed after a few seconds.

### LINKED ISSUE #1196: L.esri.FeatureLayer where clause is not always persisting
## Describe the bug

A `L.esri.FeatureLayer` where clause does not always persist when changes to it are made. Occasionally the previous where clause states of the layer reappears on the map.

This supersedes #966 at [version `2.4.0`](https://github.com/Esri/esri-leaflet/releases/tag/v2.4.0).

#996 may have had more to do with network request timings and race conditions; however, I suspect the similarly observed behavior has a different cause at version `2.4.0`.

Please also investigate #875 while researching this issue.

## Expected behavior

Using a feature layer's `setWhere` method should always persist on the map. Previously used where clauses should no longer influence what is shown on the map.

## Actual behavior

Using a feature layer's `setWhere` method does not always guarantee that the new where clause state will appear correctly on the map.

## To Reproduce

1. Go to <https://esri.github.io/esri-leaflet/examples/querying-feature-layers-1.html>.
2. Start out at the demo site with 'Any'.  Ensure all layer-specific network requests are finished.
3. Change to 'Parking'. Wait for requests to settle and finish.
4. Zoom out a bit. Again wait for requests to settle and finish.
5. Zoom back in to roughly where you started out. The 'Any' features where clause state will be shown for the layer instead of only 'Parking'.

## Screenshots

image

## Environment Information

- Version of Leaflet (`L.version`): `1.6.0`
- Version of Esri Leaflet (`L.esri.VERSION`): `2.4.0`
- Your OS: Windows
- Browser and Version: FF 76.0

## Other References

- [Geonet discussion](https://community.esri.com/thread/255032-zoom-change-causes-filter-reset)

patrickarlt:
This PR fixes the following issues with `FeatureLayer.setWhere()`

* Fix #875
* Fix #1196

#1196 is happening because the internal layer cache was not cleared when `setWhere()` was called.  #875 was caused by race conditions in calling `setWhere` in quick succession. Given that we cannot cancel requests we now simply ignore then if the `where` in the request doesn't match the current `where` of the layer.

This means that callback to `setWhere` won't be run if they get interrupted by another `setWhere()` which I think is better behavior.

Adding @gavinr and @jwasilgeo for review.

### DIFF
--- a/src/Layers/FeatureLayer/FeatureManager.js
+++ b/src/Layers/FeatureLayer/FeatureManager.js
@@ -133,7 +133,7 @@ export var FeatureManager = FeatureGrid.extend({
   _requestFeatures: function (bounds, coords, callback) {
     this._activeRequests++;
 
-    const originalWhere = this.options.where;
+    var originalWhere = this.options.where;
 
     // our first active request fires loading
     if (this._activeRequests === 1) {
--- a/src/Layers/FeatureLayer/FeatureManager.js
+++ b/src/Layers/FeatureLayer/FeatureManager.js
@@ -133,6 +133,8 @@ export var FeatureManager = FeatureGrid.extend({
   _requestFeatures: function (bounds, coords, callback) {
     this._activeRequests++;
 
+    const originalWhere = this.options.where;
+
     // our first active request fires loading
     if (this._activeRequests === 1) {
       this.fire(
@@ -151,6 +153,11 @@ export var FeatureManager = FeatureGrid.extend({
     ) {
       if (response && response.exceededTransferLimit) {
         this.fire('drawlimitexceeded');
+      }
+
+      // the where changed while this request was being run so don't it.
+      if (this.options.where !== originalWhere) {
+        return;
       }
 
       // no error, features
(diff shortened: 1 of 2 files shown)


src/Layers/FeatureLayer/FeatureManager.js
```diff
@@ -133,6 +133,8 @@ export var FeatureManager = FeatureGrid.extend({
   _requestFeatures: function (bounds, coords, callback) {
     this._activeRequests++;
 
+    const originalWhere = this.options.where;
```
gavinr, line 136:
```suggestion
    var originalWhere = this.options.where;
```

looks like the uglify plugin is not liking the es6 `const`

gavinr:
This was released in [v2.5.0](https://github.com/Esri/esri-leaflet/releases/tag/v2.5.0).

Issue #1207, opened 2020-06-23
MinZoom is not removing the layers second time around


vangenechtenbert:
Use a FeatureLayer with the minZoom option.
At first the layers act as aspected.
When we zoom in : they are added.
When we zoom back out over the minZoom level: they are removed.

Try this a second time: they are not removed.

As of version 2.4 this behavior is not working as it should.
Maby because the recent featureLayer changes?

gavinr:
@vangenechtenbert thanks for the report. Do you see any errors in the console when it stops working? Could you post a public example app that the error happens in? Thanks!

vangenechtenbert:
https://jsfiddle.net/yqrd9e3h/
Zoom in
Zoom out
Zoom in
Zoom out

The markers will no longer have effect from the minZoom option

pmacMaps:
I reviewed the Fiddler.  There was an error when the map first loaded: "Cannot read property 'extend' of undefined"; coming from Task.js.

I also noticed that sometimes the min/max zoom was honored, and other times, it was not.

![error-log-esri-leaflet](https://user-images.githubusercontent.com/12861454/85549604-283aaf80-b5ee-11ea-8ce6-a6a89e488a60.PNG)

gavinr:
Thanks! In the JSFiddle, esri-leaflet is being loaded twice - once in the script tag, and once via the jsfiddle "resources" tab:

![image](https://user-images.githubusercontent.com/209355/85557074-f853d200-b616-11ea-97ac-ad69afb900c5.png)

If you remove one of those, the "Task.js" error does not happen.

After that change, I am able to replicate the issue where the minZoom does not seem to be re-applied after the initial zoom-in/zoom-out. I don't see any console errors.

vangenechtenbert:
yep, no errors.
Just unexpected behavior :)

jwasilgeo:
Closed in PR #1212.

Issue #1193, opened 2020-04-10
Labels removed when features move outside of view and back.


freewheelCarto:
Using the code given for the [Labeling Features](https://esri.github.io/esri-leaflet/examples/labeling-features.html) example (and in the example itself), labels are sometimes removed when the features move out of the viewport and back in. 

It seems related to the removefeature events. In some cases (but not always), when the features move out of the viewport, I was able to see that the removefeature event is fired ([related to this functionality?](https://github.com/Esri/esri-leaflet/issues/408)). In the example code, the labels are removed from the map on this event. I do see that addfeature is fired when the map is panned and features make it back into view, but it doesn't add the labels. 

I'm trying to figure out a solution for my current project. If I can figure something out, it will probably be more of a work around than a permanent solution. Maybe someone will be able to address this in a more permanent way for the example on the Esri Leaflet docs.

Here is a screenshot from the Esri Leaflet example page that shows missing labels. 

![image](https://user-images.githubusercontent.com/12546498/78998922-9a374880-7b17-11ea-89e4-af23f5585798.png)

jwasilgeo:
That's an interesting catch, thanks for pointing this out.

I haven't had focused time yet to debug what is going on the way that sample was written and in your linked issue #408 comment thread, but according to the [API docs](https://esri.github.io/esri-leaflet/api-reference/layers/feature-layer.html#events) all of this _should_ work as-is.

I created a remixed codepen of the sample page in question using a different approach.  It is just a quick trial and exploration, and admittedly it may be a bit heavy-handed doing the label removing/adding on map `move` events.  Check out the console statements to see that at first glance there should be just enough label `<div>`s hanging out in the DOM for the features currently displayed.

https://codepen.io/jwasilgeo/pen/xxwbEXP

(I had used the `move` event for [a different visualization purpose in an earlier codepen](https://codepen.io/jwasilgeo/pen/PMZVLq).)

freewheelCarto:
Thanks for the quick response and quick solution, which seems like it should do the trick in the meantime.

jwasilgeo:
There's also [Geovation/labelgun](https://github.com/Geovation/labelgun). Would be interesting to compare/contrast and give it a whirl. They've a live demo with esri-leaflet and [here's the source](https://github.com/Geovation/labelgun/blob/master/docs/examples/esri-leaflet/index.html).

Ah, and there's an existing conversation that touches on labelgun over in #916.

jwasilgeo:
[...]

PR #1212, opened 2020-07-09
do not clear _currentSnapshot when hiding the layer

gavinr:
The behavior in #1207 is happening because the first time the layer is hidden due to the maxZoom, [`this._currentSnapshot` is emptied](https://github.com/Esri/esri-leaflet/blob/0e89c5e12e4387a22510a4bee628126298482413/src/Layers/FeatureLayer/FeatureManager.js#L505), so the [next time it tries to hide the layers](https://github.com/Esri/esri-leaflet/blob/0e89c5e12e4387a22510a4bee628126298482413/src/Layers/FeatureLayer/FeatureManager.js#L504), `this._currentSnapshot` is empty so it does not remove anything.

The bug is fixed if you remove that line (this PR), but I'm not quite convinced this is the correct fix because I'm guessing there's some side-effect that requires that line, and I'm just not understanding it - @patrickarlt @jgravois please let me know if you have any insights into why this line (`this._currentSnapshot = [];`) was [originally included](https://github.com/Esri/esri-leaflet/commit/4bffb46e44fe1e809645361299a40f16a4d483ec#diff-b59294fc1c59924a39c0eb1607d4cc9cR74).

bug fixed

### DIFF
--- a/src/Layers/FeatureLayer/FeatureGrid.js
+++ b/src/Layers/FeatureLayer/FeatureGrid.js
@@ -388,7 +388,9 @@ export var FeatureGrid = Layer.extend({
       }
 
       for (i = 0; i < queue.length; i++) {
-        if (this._activeCells[this._cellCoordsToKey(queue[i])]) {
+        var _key = this._cellCoordsToKey(queue[i]);
+        var _coords = this._keyToCellCoords(_key);
+        if (this._activeCells[_coords]) {
           this._reuseCell(queue[i]);
         } else {
           this._createCell(queue[i]);
--- a/src/Layers/FeatureLayer/FeatureManager.js
+++ b/src/Layers/FeatureLayer/FeatureManager.js
@@ -502,7 +502,6 @@ export var FeatureManager = FeatureGrid.extend({
   _handleZoomChange: function () {
     if (!this._visibleZoom()) {
       this.removeLayers(this._currentSnapshot);
-      this._currentSnapshot = [];
     } else {
       /*
       for every cell in this._cells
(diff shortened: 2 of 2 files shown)


jgravois:
🤷 nobody has touched that code in 5 years. 

Screen Shot 2020-07-09 at 3 25 32 PM

off the top of my head it doesn't seem problematic to preserve the cache instead of destroying it in the block. 

the only other thing i'd suggest testing would be applying a new filter both while the layer is at a visible zoom level and when its not. i have a very vague recollection of folks complaining about a stale cache in a scenario like that.

gavinr:
Thanks @jgravois - you were correct there was an issue when zooming in, calling `setWhere`, then zooming out/in. \

Seems like in `_requestFeatures` function second parameter should be an object (correct [here](https://github.com/Esri/esri-leaflet/blob/0e89c5e12e4387a22510a4bee628126298482413/src/Layers/FeatureLayer/FeatureManager.js#L129)), but in three other places it was sending a string which was causing issues. Fixed that in 87b4d14.

patrickarlt:
@gavinr this looks like a good change. Lets merge this.

gavinr:
thanks Patrick. @jwasilgeo found one small issue - i'm going to try to fix that quickly before merge.

jwasilgeo:
@gavinr, I'd like to also add the [Labeling Features demo page](https://esri.github.io/esri-leaflet/examples/labeling-features.html) to our list of manual tests. The odd behavior described in https://github.com/Esri/esri-leaflet/issues/1193 was fixed as a byproduct of `v2.4.0` and I'd like to close it out if it still works as expected with the changes in this PR.

Issue #1222, opened 2020-07-28
Dynamically resample Esri World Imagery


jwasilgeo:
Investigate the possibility of dynamically resampling the Esri World Imagery raster basemap which has tiles available in some regions of the world beyond zoom level 19.

**Additional context**

- Original issue #240.
- PR #1011 (it resolved #240 but we decided to revert on 27 July 2020; see discussion in issue #1210).
- Issue #1210 where we identified some behavior that needed extra attention to fix.
    - Need to understand how to deal with the missing tiles behavior described in https://github.com/Esri/esri-leaflet/issues/1210#issuecomment-662691014, https://github.com/Esri/esri-leaflet/issues/1210#issuecomment-662772699, and https://github.com/Esri/esri-leaflet/issues/1210#issuecomment-663065941.
- Take care not to attempt to apply these changes to other imagery-related basemaps (see #1197).

PR #1223, opened 2020-07-28
revert-pr-1011

### LINKED ISSUE #1210: The maxNativeZoom option of the layer is not being set when using L.esri.basemapLayer
## Bug
The `maxNativeZoom` option of the layer is not being set when using the method `L.esri.basemapLayer('Streets')` along with the options object. This means that some parts of the map display the "Map data not yet available" when the zoom level is greater than 19. 

I've tried two different ways of setting the maxZoomLevel, both unsuccessful:

1.
```
var map = L.map('map');
L.esri.basemapLayer('Imagery', { maxNativeZoom: 19}).addTo(map);
```

2. 
```
var map = L.map('map');
var layer = L.esri.basemapLayer('Imagery');
layer.options.maxNativeZoom = 19;
layer.addTo(map)
```

## Expected Behavior
The map should be auto-scaled when zooming over the set `maxNativeZoom` preventing the "Map data not yet available" tiles from being displayed. Using the `L.tileLayer` Leaflet method with the appropriate Url for Imagery and setting the `maxNativeZoom` option does result in the expected behavior.

## Screenshots

image

## Environment Information

- Version of Leaflet (`L.version`): 1.6.0
- Version of Esri Leaflet (`L.esri.VERSION`): 2.3.2
- Your OS: Windows 10
- Browser and Version: Chrome 83.0.4103.116

jwasilgeo:
reverting changes introduced in pr #1011

resolves #1210 

see enhancement issue #1222 for revisiting bringing this back in down the road

### DIFF
--- a/src/Layers/BasemapLayer.js
+++ b/src/Layers/BasemapLayer.js
@@ -1,6 +1,5 @@
 import { TileLayer, Util } from 'leaflet';
 import { pointerEvents } from '../Support';
-import { request } from '../Request';
 import {
   setEsriAttribution,
   _getAttributionData,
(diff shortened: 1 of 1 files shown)


gavinr:
The code changes look great, **_but_** ....

I noticed that after these changes, if you zoom into the Atlantic ocean, you get gray "map data not yet available" errors starting at zoom level 14: http://somup.com/cYiuD262nw .. I thought we were expecting that if we reverted this and set the maxZoom to 19, then we would at least have tiles everywhere on the world.

gavinr:
Huh, I never noticed that. Given that it's now consistent with the other basemaps, sounds good to me to merge this.

Issue #665, opened 2015-10-23
Layer refresh not removing unwanted markers


Greigrm:
My use case here may not be valid, so please close if not relevant.

I would like a layer where I can effectively leave the map screen open and have the layer re-queried in the background and any added markers placed onto the map, and any missing markers removed.  I'm not so concerned about changes to items staying on the map, just really the add/delete.

Looking at the disussion for #533 I have tried to use the featureLayer.refresh() method on a setTimeout, and while this adds new markers, it does not remove markers which are not in the result set.

To work around this I then tried simply calling the .setWhere() method (passing the existing where without change) and this works better and correctly removes deleted items, however if a popup is opened it will close when the setWhere() is triggered.

So I'm not sure if this is a bug report for refresh() not deleting markers, or a feature request for it to do the deletes.

jgravois:
`L.esri.featureLayer.refresh()` should definitely fetch newly created features *and* remove features from the map that are no longer present in the service.

I'll check into it.

patrickarlt:
@jgravois assigning to you. Take a look at how I did `setWhere` https://github.com/Esri/esri-leaflet/blob/master/src/Layers/FeatureLayer/FeatureManager.js#L180-L225.

The problem we are going to have is that right now `setWhere` destroys all the existing layers and then re-adds everything which is why the popup is closing when @Greigrm calls `setWhere` with the same where. It is definitely non-trivial to figure out that feature "X" is already active and does/doesn't need to be removed but you could definitely give it a shot.

jgravois:
:+1:

gavinr:
It looks like this is still an issue. Reproduction steps:

1. https://jsbin.com/kutivirogi/1/edit?html,output
2. Click on the map a few times to create some points.
3. Click "delete all" button
4. Click "refreshLayer" button
    - Expected: all points on the map disappear
    - Actual: all points stay there

gavinr:
This was fixed in [v2.5.0](https://github.com/Esri/esri-leaflet/releases/tag/v2.5.0).

PR #1224, opened 2020-07-28
refresh use setWhere

gavinr:
trying to address #665, taking advantage of the code in `setWhere()` - there may be a catch here that I'm missing though, would love your input @patrickarlt @jgravois 

<details>
 <summary>sample.html</summary>

```html
<!DOCTYPE html>
<html>

<head>
  <meta charset="utf-8" />
  <title>Editing feature layers</title>
  <meta name="viewport" content="initial-scale=1,maximum-scale=1,user-scalable=no" />

  <!-- Load Leaflet from CDN -->
  <link rel="stylesheet" href="https://unpkg.com/leaflet@1.6.0/dist/leaflet.css"
    integrity="sha512-xwE/Az9zrjBIphAcBb3F6JVqxf46+CDLwfLMHloNu6KEQCAWi6HcDUbeOfBIptF7tcCzusKFjFw2yuvEpDL9wQ=="
    crossorigin="" />
  <script src="https://unpkg.com/leaflet@1.6.0/dist/leaflet.js"
    integrity="sha512-gZwIG9x3wUXg2hdXF6+rVkLF/0Vi9U8D2Ntg4Ga5I5BZpVkVxlJWbSQtXPSiUTtC0TjtGOmxa1AJPuV0CPthew=="
    crossorigin=""></script>

  <script src="../dist/esri-leaflet-debug.js"></script>

  <!-- Load Esri Leaflet Renderers plugin to use feature service symbology -->
  <script src="https://unpkg.com/esri-leaflet-renderers@2.1.2"></script>

  <style>
    body {
      margin: 0;
      padding: 0;
    }

    #map {
      position: absolute;
      top: 0;
      bottom: 0;
      right: 0;
      left: 0;
    }
  </style>
</head>

<body>

  <!-- Leaflet Draw -->
  <style>
    #info-pane {
      position: absolute;
      top: 10px;
      right: 10px;
      z-index: 400;
      padding: 1em;
      background: white;
      text-align: right;
      max-width: 250px;
    }

    #form {
      display: none;
    }
  </style>

  <div id="map"></div>
  <div id="info-pane" class="leaflet-bar">
    <label>
      Click on the map to post an edit.<br>
      Click on a marker to remove it.
      <select id="symbols">
        <option value="Air Control Point">Air Control Point</option>
        <option value="Decision Point">Decision Point</option>
        <option value="Unmanned Aerial System (UAS UA)">Unmanned Aerial System (UAS UA)</option>
      </select>
    </label>

    <button id="refreshLayer">refreshLayer</button><br />
    <button id="deleteAllAndCreate">delete all</button>
  </div>

  <script>
    // create the map
    var map = L.map('map').setView([65, -18.5], 14);
    L.esri.basemapLayer('Physical').addTo(map);

    // add our feature layer to the map
    const militaryOps = L.esri.featureLayer({
      url: 'https://sampleserver6.arcgisonline.com/arcgis/rest/services/Military/FeatureServer/3'
    }).addTo(map);

    var symbols = document.querySelector('#symbols');

    // when the map is clicked, add a new feature
    map.on('click', function (e) {
      // convert to GeoJSON
      const feat = L.marker(e.latlng).toGeoJSON();
      // set the attribute value that controls the feature service symbology
      feat.properties.symbolname = symbols.value;
      // make a request to add the new feature to the feature service.
      militaryOps.addFeature(feat, function (err, response) {
        if (err) {
          return;
        }

        console.log(response);
      });
    });

    // when a marker is clicked, remove the feature from the service, using its id
    militaryOps.on('click', function (e) {
      militaryOps.deleteFeature(e.layer.feature.id, function (err, response) {
        if (err) {
          return;
        }

        console.log(response);
      });
      // make sure map click event isn't fired.
      L.DomEvent.stopPropagation(e);
    });

    document.getElementById('refreshLayer').addEventListener('click', (evt) => {
      console.log('REFRESH BUTTON CLICK',);
      militaryOps.refresh();
    });

    document.getElementById('deleteAllAndCreate').addEventListener('click', (evt) => {
      L.esri.post('https://sampleserver6.arcgisonline.com/arcgis/rest/services/Military/FeatureServer/3/deleteFeatures', {
        where: '1=1'
      }, function (error, response) {
        if (error) {
          console.log(error);
        } else {
          console.log(response.name);
        }
      });
    })
  </script>

</body>

</html>
```
</details>

### DIFF
--- a/src/Layers/FeatureLayer/FeatureManager.js
+++ b/src/Layers/FeatureLayer/FeatureManager.js
@@ -361,23 +361,7 @@ export var FeatureManager = FeatureGrid.extend({
   },
 
   refresh: function () {
-    for (var key in this._cells) {
-      var coords = this._keyToCellCoords(key);
-      var bounds = this._cellCoordsToBounds(coords);
-      this._requestFeatures(bounds, coords);
-    }
-
-    if (this.redraw) {
-      this.once(
-        'load',
-        function () {
-          this.eachFeature(function (layer) {
-            this._redraw(layer.feature.id);
-          }, this);
-        },
-        this
-      );
-    }
+    this.setWhere(this.options.where);
   },
 
   _filterExistingFeatures: function (oldFrom, oldTo, newFrom, newTo) {


jgravois:
> if a popup is opened it will close when the setWhere() is triggered.

is ☝️ still an issue? ([ref](https://github.com/Esri/esri-leaflet/issues/665#issue-113095882))

gavinr:
Looks like it does not close the popup:
http://somup.com/cYiUDP6okM

... which is preferable according to the original issue.

In the example in the recording (where the feature was removed), however, it would probably be preferable to close it. The preference on what you'd want it to do depends on the specific edits that have happened between the original layer load and the refresh I guess.

jgravois:
awesome. great job tackling an old chestnut. 🔩 

> depends on the specific edits that have happened between the original layer load and the refresh I guess.

agreed. i wouldn't expect the library to be clever enough to close the popup because the feature was removed upstream.

jwasilgeo:
I like the way these changes are headed in relying on `setWhere`, but I was able to still get invalid features to show up on the map. The exact repro steps aren't clear yet, but while using your sample code above (thanks for that, btw) with some combo of panning away from the initial extent and/or zooming out, then adding new features here and there like a salt shaker, and finally pressing 'delete' followed by 'refresh' one can occasionally get incorrect fossil features hanging out on the map.

gavinr:
I think I discovered the issue: the `refresh()` call does not delete points that have been added since the page load (no panning needed)

gavinr:
@jwasilgeo I think we have discovered a separate issue specifically around editing in esri-leaflet and the `setWhere` functionality. 

If you take the editing out of the picture by using this app to do all editing:
https://arcgis.com/home/webmap/viewer.html?webmap=21e58350f6884ca49505a94ec538ccec
(do all editing there and then just "refreshLayer" button in the esri-leaflet debug app to see if the changes are applied)
.... it seems to work correctly for me.

jwasilgeo:
Thanks for the additional debugging and clear info, @gavinr. I agree with and can verify what you've uncovered.

PR #1233, opened 2020-10-02
Support 'withCredentials' option which will include credentials with CORS requests

luiscamachopt:
addresses #881

### DIFF
--- a/spec/Layers/RasterLayerSpec.js
+++ b/spec/Layers/RasterLayerSpec.js
@@ -10,6 +10,7 @@ describe('L.esri.RasterLayer', function () {
   // Extend 'abstract' RasterLayer object and implement functionality required for tests
   var TestRasterLayer = L.esri.RasterLayer.extend({
     initialize: function (options) {
+      L.setOptions(this, options);
       this.service = {
         metadata: function () {}
       };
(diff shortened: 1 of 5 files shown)


gavinr:
@luiscamachopt thank you for the PR! A few small requests:

1. It looks like the unit tests are failing, mostly linting issues. Please run "npm test" locally and fix any linting issues.
1. Issue 881 is quite long, could you please add a summary/description here of the code changes you're proposing here, why they're needed, and how to see the changes in action (replication case)? 
1. Could you please add some unit tests that will test this behavior to ensure it works going into the future.

Thanks!

luiscamachopt:
When sending cross origin requests to servers that are secured with Active Directory (issue refers to Integrated 'Windows Authentication' IWA)  it is necessary to send http requests with 'withCredentials' set to true and image elements should property 'crossOrigin' set to 'use-credentials'
With this PR, If the layer is created with the option 'withCredentials' set to true the requests will be adjusted as described so that the remote server receives the credentials and accepts the requests.

luiscamachopt:
Hi @gavinr
I've fixed the linting issues, pushed tests and added a more complete description of the issue.
Thanks

patrickarlt (approved):
From a code perspective this looks good. However this change is difficult for us to test given that:

1. The original issue is old.
2. The original group debugging the issue has generally moved on.
3. We don't have servers readily available to test this setup.

I'm OK with merging but @gavinr can you do some extensive testing against our existing examples to ensure their are no regressions.

gavinr:
Tested against all existing samples, looks ok. Merging.

gavinr:
This was released in [v2.5.1](https://github.com/Esri/esri-leaflet/releases/tag/v2.5.1) - thanks!

PR #1235, opened 2020-10-05
use rollup-plugin-terser instead of rollup-plugin-uglify

gavinr:
Right now there's a security issue:

```
High            Remote Code Execution
  Package         serialize-javascript
  Patched in      >=3.1.0
  Dependency of   rollup-plugin-uglify [dev]
  Path            rollup-plugin-uglify > serialize-javascript
  More info       https://npmjs.com/advisories/1548
found 1 high severity vulnerability in 832 scanned packages
```

It seems like this plugin [might not be supported anymore](https://github.com/TrySound/rollup-plugin-uglify/pull/85#issuecomment-674202921), (https://github.com/TrySound/rollup-plugin-uglify/issues/87) so switching this to an alternate. It seems like the copyright note is not on the output `esri-leaflet.js` file though - may need to look into this.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3195,22 +3195,29 @@
       }
     },
     "jest-worker": {
-      "version": "24.9.0",
-      "resolved": "https://registry.npmjs.org/jest-worker/-/jest-worker-24.9.0.tgz",
-      "integrity": "sha512-51PE4haMSXcHohnSMdM42anbvZANYTqMrr52tVKPqqsPJMzoP6FYYDVqahX/HrAoKEKz3uUPzSvKs9A3qR4iVw==",
+      "version": "26.3.0",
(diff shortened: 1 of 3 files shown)


jwasilgeo:
If this change goes through, then I suggest we also think about updating these plugins' npm deps that rely on `esri-leaflet`'s own Rollup config:
- https://github.com/Esri/esri-leaflet-vector
- https://github.com/Esri/esri-leaflet-geocoder
- https://github.com/Esri/esri-leaflet-cluster
- https://github.com/Esri/esri-leaflet-heatmap

This plugin does its own Rollup config, but we should see if anything needs to be done:
- https://github.com/Esri/esri-leaflet-renderers

Am I missing anything?

gavinr:
Closing in favor of #1315

Issue #1239, opened 2020-11-06
"Powered by Esri" Map Attribution doesn't go away when you run Map.removeLayer(...) on an esri-leaflet map


timothycouchhsi:
## Describe the bug

If you add and then later remove an esri-leaflet layer, the "Powered by Esri" attribution stays on the map.

## Expected behavior

I'm trying to make a toggleable esri-leaflet layer. I would like to have Powered by Esri come up when I toggle one of these layers on, then I would like it to remove from the attributions when I toggle the layer back off. 

## Actual behavior

I'm using L.Map.removeLayer to remove my esri-leaflet map layer, but the "Powered by Esri" attribution stays on the attribution even when the layer is gone. It does not create multiple attributions if I toggle the layer many times or have multiple esri-leaflet layers at once.

## Screenshots

![image](https://user-images.githubusercontent.com/29982559/98412891-591bfe00-203e-11eb-83c1-fb44d265a633.png)

## Environment Information

- Version of Leaflet - 1.3.1
- Version of Esri Leaflet 2.3.3, 2.5.1
- Your OS: Windows 10
- Browser and Version Chrome 86.0.4240.183 64-bit

## Additional context

Using webpack.

Thank you!

gavinr:
Hi, thanks for posting this issue. What kind of layer are you seeing this behavior with (Basemap Layer, Feature Layer, etc)? 

Do you have a jsbin that demonstrates the issue?

Thanks!

timothycouchhsi:
I'm using a dynamicMapLayer and a tiledMapLayer for base layers. Sorry, but I don't have a jsbin to show. Thank you for checking this out!

jwasilgeo:
Thanks for the additional info but a live public demo would be extremely useful and helpful. If not JSBin, perhaps you could us <https://codesandbox.io> or <https://codepen.io/pen/> or provide a full html/js/css code sample re-using some public layer resources available at the demo pages <https://esri.github.io/esri-leaflet/examples/>?

pjbiogit:
I have also run into this problem. You can see a simple demonstration [here](https://codepen.io/pjbiogit/pen/RwGLaaN). After adding the layer (via button) the attribution is added to the map, but then when removing the layer (pressing the other layer) the attribution is not removed. 

[...]

PR #1248, opened 2020-12-28
remove attribution logic

gavinr:
This is a follow-up to #841 where we added the logic to _add_ the attribution, but did not add the logic to _remove_ the attribution string:
- Track how many esri-leaflet layers are on the map (stored as in integer in `attributionControl._esriAttributionLayerCount`)
- using the count from `attributionControl._esriAttributionLayerCount`, remove the "Powered by Esri" attribution when all the esri-leaflet layers have been removed from the map

#1239

### DIFF
--- a/src/Layers/BasemapLayer.js
+++ b/src/Layers/BasemapLayer.js
@@ -2,6 +2,7 @@ import { TileLayer, Util } from 'leaflet';
 import { pointerEvents } from '../Support';
 import {
   setEsriAttribution,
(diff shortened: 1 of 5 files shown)


jwasilgeo (approved):

jwasilgeo:
looks great!

gavinr:
Released in https://github.com/Esri/esri-leaflet/releases/tag/v2.5.2

Issue #1250, opened 2020-12-31
Nearby queries incorrectly sending param "inSr" instead of "inSR"


Voileexperiments:
I'm debugging a issue where a `nearby` query on a ArcGIS Enterprise feature layer does not work at all if it's initiated from leaflet (it either returns full result or empty), but works completely fine when testing on REST API page. After some squinting I discovered the only difference between two requests is that ArcGIS REST API page send `inSR=4326` while Leaflet sends `inSr=4326`, which caused the `nearby` query to fail.

[The correct param](https://developers.arcgis.com/rest/services-reference/query-feature-service-layer-.htm) is `inSR`. Using `inSr` works for ArcGIS Online (https://esri.github.io/esri-leaflet/examples/querying-feature-layers-2.html) because presumably it's more lax with request params case-sensitivity.

I think there are a total of two places that needs to be changed, and the fix should be easy(?):

https://github.com/Esri/esri-leaflet/blob/master/src/Tasks/Query.js#L97
https://github.com/Esri/esri-leaflet/blob/master/src/Tasks/Query.js#L213

jwasilgeo:
@Voileexperiments thank you for the clear description and doing some research on what's going on. We'll take a look.

jwasilgeo:
@Voileexperiments can you also tell us what version of ArcGIS Enterprise you are seeing this issue with?

jwasilgeo:
Related to PR #1168 and its issue #1167 which focused on the `outSR` param.

jwasilgeo:
This will be available the next time we publish a new release.

PR #1251, opened 2020-12-31
1250-inSR

jwasilgeo:
Changed:

- REST calls will now use `inSR` instead of `inSr`

#1250


gavinr (approved):
Looks great!

(to reiterate, this is basically the equivalent of #1168, but addressing *in*SR this time)

gavinr:
This has been released in [v2.5.3](https://github.com/Esri/esri-leaflet/releases/tag/v2.5.3).

PR #1252, opened 2021-01-05
Only attempt to set/remove attribution if control exists

avinmathew:
There were multiple places were the existence of the attribution control was checked, but for some code paths, it wasn't checked causing exceptions to be thrown.

Use a single check for the existence of the attribution control at the beginning of setting/removing attribution.

### DIFF
--- a/src/Util.js
+++ b/src/Util.js
@@ -174,13 +174,17 @@ export function calcAttributionWidth (map) {
 }
 
 export function setEsriAttribution (map) {
+  if (!map.attributionControl) {
+    return;
+  }
+
   if (!map.attributionControl._esriAttributionLayerCount) {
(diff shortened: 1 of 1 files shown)


gavinr:
It looks like before #1248, if you created a map with `attributionControl` to `false`, it work work fine:

```
var map = L.map('map', {
    attributionControl: false
  }).setView([37.837, -122.479], 8);

  var basemapLayer = L.esri.basemapLayer('Streets');
  basemapLayer.addTo(map);
```

... but after #1248 that case now errors. So this appears to be a regression - sorry about that, and thanks for the PR!

gavinr:
This has been released in [v2.5.3](https://github.com/Esri/esri-leaflet/releases/tag/v2.5.3).

Issue #749, opened 2016-02-23
Cancel Previous Pending Requests


bdk0172:
We are displaying a customer's ArcGIS tiles within Mapbox and every time zoom is used, requests for tiles are sent out.  With this particular customer, not sure if their server is slow or experiencing network issues, but it appears that it gets hung up and all requests are sitting in a pending status when looking at the network tab in Chrome developer tools.  Is there anyway possible to cancel previous requests when another zoom is done?  It has to wait for all previous requests to return before it displays the tiles for the final zoom level.  In some cases this is taking a minute plus.

bbehling:
I've seen similar issues using ArcGIS sever tiles and dynamic services.

One issue is where the last pending request is drawn on the map, even after calling the setLayers() method with no layers turned on. So, we end up with tiles drawn when actually there shouldn't be any map tiles present on the map.

Another issue is in Internet Explorer 11. All requests are stuck in a pending state and never draw on the map, but that is occurring because I had have to use the es6-promise polyfill to fix issues with bing maps plugins.

Lastly, I've seen issues where tiles will not draw in the correct on the map. We consume multiple services, and those are placed in the map in a particular order. If one of those services takes longer to finish a request then other services, that particular service is rendered underneath the services that completed first, when what should happen is all services should be rendered in the same order they were added to the map, regardless of when the requests finished. 

It seems request promises are not honoring the order the services are in, or just issues with handling promises in general? All these asynchronous request issues are happening with or without the es6 promise polyfill.

I have a demo app using the ESRI Leaflet API which I ported over from an ESRI JSAPI project. Using the same logic to toggle layers, I've never had a problem with pending requests or services drawn out of order.

jgravois:
> what should happen is all services should be rendered in the same order they were added to the map, regardless of when the requests finished.

that isn't how things work in Leaflet.  if you want to get explicit control over the drawing order in Leaflet, you'll need to use [custom panes](http://esri.github.io/esri-leaflet/examples/layer-ordering.html).

bbehling:
Custom panes applies to all types of layers, not just feature layers?

jgravois:
yup. they can even be used to draw images on top of graphics.

bbehling:
Thanks.

Any thoughts on the canceling pending request issue?

jgravois:
do you have a repro case you can share?

bbehling:
Custompanes - 

[...]

PR #1261, opened 2021-02-11
Disregard stale DynamicMapLayer updates

davetapley:
Alleviates #749  by not updating layer with image for a set of params which have changed while request was in flight.

This isn't the most elegant solution; I'm very open to feedback 😁


jwasilgeo:
Hi @davetapley, thank you for offering this change, we appreciate it! 👍 

As a holistic, longer-term update, I'd like to see esri-leaflet's requests mechanisms be updated to support network requests that can be aborted/canceled, as folks have mentioned in #749. It is on our radar to evaluate and consider that more closely, as it would be a bit of an update that would likely lead to a major version release.

I think for now we'll leave your proposed change parked here while we figure out next steps.

AStasyK:
Hello @jwasilgeo, We're having the same issue in our application. I hope the solution is progressing. It will be very important!

Issue #1249, opened 2020-12-30
warning with webpack5


wwwwlm-work:
## Describe the bug
not ok when using webpack5.

## Expected behavior
should use default export instead of named export

## Actual behavior
there comes a warning as blow:
in /node_modules/esri-leaflet/src/EsriLeaflet.js
Should not import the named export 'version' (reexported as 'VERSION') from default-exporting module (only default export is available soon) 
![image](https://user-images.githubusercontent.com/73639742/103335433-de0c1c80-4aaf-11eb-802e-692468e24eb1.png)

## To Reproduce
using webpack5

## Environment Information
- Version of Leaflet (`L.version`): 1.7.1
- Version of Esri Leaflet (`L.esri.VERSION`): 2.5.2
- Your OS: window 10
- Browser and Version Chrome 87.0.4280.88

## Additional context
webpack: 5.11.0;

gavinr:
Thanks for the report - could you please post more specific replication steps?

jgravois:
@gavinr is right. a dummy application with the bare minimum code needed to reproduce the error would help the maintainers here focus on solving the problem instead of repro-ing the behavior you're reporting.

based on the information @wwwwlm-work already provided, its possible webpack@5 is finding fault with the current mix of named and [default exports](https://github.com/Esri/esri-leaflet/search?q=export+default).

https://github.com/Esri/esri-leaflet/blob/347397989212f3f80d5360163f6dd7062f10670b/src/Request.js#L248

The rollup doc confirms that the practice is frowned upon: https://rollupjs.org/guide/en/#default-export

> It is bad practice to mix default and named exports in the same module, though it is allowed by the specification.

i haven't given it much thought, but i'd expect the appropriate action to be _getting rid_ of the `default` exports, not the other way around.

wwwwlm-work:
here is the demo repo to reproduce : https://github.com/wwwwlm-work/issue-demo

blow is what I did to prevent that warning temporarily:
```javascript
// current
export {version as VERSION} from '../package.json';

// improved
import pkg from '../package.json'
const VERSION = pkg.version;
export { VERSION };
```

Happytoad42:
Any update on this issue?

daanielcarvalho:
Hello

There is any solution on this? 

```
ERROR in ./node_modules/esri-leaflet/src/EsriLeaflet.js 2:0-51
Should not import the named export 'version' (reexported as 'VERSION') from default-exporting module (only default export is available soon)
[...]

PR #1273, opened 2021-05-24
fix: Use default export from package instead of named export version

destus90:
https://webpack.js.org/migrate/5/#cleanup-the-code

> Using named exports from JSON modules: this is not supported by the new specification and you will get a warning. Instead of import { version } from './package.json' use import package from './package.json'; const { version } = package;

With this fix allow to use Webpack 5 with no warning emitted.

Fixes https://github.com/Esri/esri-leaflet/issues/1249

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased][unreleased]
 
+### Fixed
+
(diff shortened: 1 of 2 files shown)


gavinr (approved):
looks good to me - thanks for the PR!

jwasilgeo (approved):

gavinr:
released in v3.0.2

PR #1288, opened 2021-08-11
updated sample code and issue template

gavinr:
The sample code in the README was using a service that does not exist anymore. I found a similar service (national parks) that seems to be well maintained (https://www.arcgis.com/home/item.html?id=8b98df1300b749eabc3142864bb9a119) and switched to that. This code is what is in the JSBin link in our Issue Template, so I also created a new JSBin template and linked to that from the issue template.


jwasilgeo (approved):
Looks great, thank you!

PR #1301, opened 2021-12-01
fix #1289 and fix #1298

### LINKED ISSUE #1289: eachActiveFeature doesn't respect minZoom/maxZoom parameters
jsbin: https://jsbin.com/ketilafebo/edit?html,console,output

There is a feature layer set to `minZoom = 7` while the map has initial `zoom = 6`, so the features in the feature layer are not loaded and there are no active features:

```
var features = [];
parks.eachActiveFeature(layer => features.push(layer));
console.log(features.length); // 0
```

When I zoom in once, the map now has `zoom = 7` so the features are loaded, as expected:

```
var features = [];
parks.eachActiveFeature(layer => features.push(layer));
console.log(features.length); // 42
```

When I zoom out again, `zoom = 6` again and the features are not shown on the map, as expected. However these features still show up in `eachActiveFeature`:

```
var features = [];
parks.eachActiveFeature(layer => features.push(layer));
console.log(features.length); // 69
```

This contradicts what `eachActiveFeature` is supposed to do:

> Calls the passed function against every feature that is currently being displayed.

(I only used `minZoom` in the jsbin but I'd assume `maxZoom` has the same problem. Also there are other filter fields in `featureLayer`, like `from`/`to`. Also `cacheLayers` is *not* considered a solution here.)

### LINKED ISSUE #1298: `addfeature` and `removefeature` events are called too often since v2.4.0 when using minZoom/maxZoom
## Describe the bug

Since Esri Leaflet v2.4.0, the `addfeature` and `removefeature` events are not working properly on a FeatureLayer with  a minZoom/maxZoom.

## Expected behavior

When there is a `minZoom` or `maxZoom` defined on the layer, the `addfeature` and `removefeature` events should get called as the user zooms "when a feature on the layer is removed from the map" (`removefeature`) and "when a previously removed feature is added back to the map" (`addfeature`) (quotes from the [docs](https://esri.github.io/esri-leaflet/api-reference/layers/feature-layer.html#events))

## Actual behavior

As you zoom out and in, the `addfeature` and `removefeature` events are fired too many times.

## To Reproduce

Esri Leaflet v2.3.3 - working as expected: https://jsbin.com/duyupof/edit?html,output
1. Open the developer tools
2. Zoom out 3-4 times
3. Notice that  "remove feature" events are only fired once when the points disappear (expected).

Esri Leaflet v3.0.3 (latest) - NOT working as expected: https://jsbin.com/yoyifa/2/edit?html,output
1. Open the developer tools
2. Zoom out 3-4 times
3. Notice that "remove feature" events are fired multiple times. Expected: Only fire once.

## Environment Information

- Version of Leaflet (`L.version`): 1.7.1
- Version of Esri Leaflet (`L.esri.VERSION`): happening on > v2.4.0

## Additional context

1. I think this behavior started in v2.4.0, from PR #1189.
2. A more specific version of this bug was originally reported in #1297.
3. Potential solution: I think we need to move or change the code where we are currently firing the events ([here](https://github.com/Esri/esri-leaflet/blob/d09bfe4b7d39770967468c9c21bf33b7773c79f6/src/Layers/FeatureLayer/FeatureLayer.js#L97-L99) and [here](https://github.com/Esri/esri-leaflet/blob/d09bfe4b7d39770967468c9c21bf33b7773c79f6/src/Layers/FeatureLayer/FeatureLayer.js#L128-L130))

patrickarlt:
This fixes both #1289 and #1298. Specifically:

* The internal `_currentSnapshot` which is a snapshot of all layers the feature layer has on the map was't properly being cleared when the layer went outside the zoom range. This is the root cause of both issues since it is used in `eachActiveFeature` and to determine which feature `removefeature` should fire for.
* There is also a missing `addfeature` event when previously removed layers were added back to the map.

https://github.com/Esri/esri-leaflet/compare/fix-feature-layer-regressions?expand=1#diff-775f68f7ed2cadfe77df5f560b04512990231859cb3e0cfee6a8c79f305eb54aR523

https://github.com/Esri/esri-leaflet/compare/fix-feature-layer-regressions?expand=1#diff-1d3459775056d23611b051bf252a9d1f457744484695c2ac7ada4f0390f0451dR189-R195


gavinr (approved):
looks good!

jwasilgeo (approved):
I agree, looks good, thanks @patrickarlt

PR #1302, opened 2021-12-02
Fix #1228 safari partially rendering

### LINKED ISSUE #1228: Partial Drawing of dynamicMapLayer in Safari
## Describe the bug

I have a list of layers that users can turn on and off at will, a user is reporting that they are only ever able to see half of a certain layer drawing when turned on.  This is the layer in question: https://location.cabarruscounty.us/arcgisservices/rest/services/DataExplorerDatasets/MapServer/1.  

When the user pans or zooms half of the screen updates while the other half remains undrawn.  I have `map.invalidateSize` setup to refresh when user resizes the screen and what not but seems that the drawing issue persists and **only does so** in Safari version 13.1.  All other browsers seem to be fine as I cannot reproduce the issue

## Screenshots

image
image

## Environment Information

- Version of Leaflet (`L.version`): 1.6.0

- Version of Esri Leaflet (`L.esri.VERSION`): 2.4.1
- Browser and Version: Safari 13.1

## Additional context

Also be aware that the layer in question is an annotation layer.

nicolaiwolko:
### Problem
See issue 1228 (https://github.com/Esri/esri-leaflet/issues/1228)
Safari sometimes only partially renders the image layer. 

### Solution
The bug is related to the leaflet library. 
Safari seems to have problems decoding too many images in main thread. 
I would suggest leaflet maintainers to switch from img to canvas to avoid such decoding problems.

In the meantime setting the img.decoding attribute to "async" will fix the bug and ensures full rendering of the layer image.

Stackblitz with bugfix: https://stackblitz.com/edit/web-platform-dmw9yg?file=esri-leaflet-debug.js

### DIFF
--- a/src/Layers/RasterLayer.js
+++ b/src/Layers/RasterLayer.js
@@ -209,6 +209,11 @@ export var RasterLayer = Layer.extend({
         interactive: this.options.interactive
       }).addTo(this._map);
 
+      // See issue 1228 (https://github.com/Esri/esri-leaflet/issues/1228)
+      // Safari sometimes only partially renders imgs with decoding = sync (safari default)
+      // To enforce full rendering of img use decoding = async
+      image._image.decoding = 'async';
+
       var onOverlayError = function () {
         this._map.removeLayer(image);
         this.fire('error');


gavinr:
Thank you for investigating and for the PR! We really appreciate your time.

In the stackblitz example, I'm noticing a "flash" of the layer when I zoom in. Here's a video (in the video it only happens once because the frame rate of the recording - it does actually happen on every zoom for me):

https://user-images.githubusercontent.com/209355/144464145-536b7a5f-ca1c-4450-b963-58dd26c51cd1.mp4

(happening me on Windows 10, Chrome v96.0.4664.45)

Is this happening for you too?

gavinr:
Inactive - closing.

PR #1308, opened 2022-02-01
added siteData.json generator

gavinr:
This is a script to generate `siteData.json`. The script is similar to the one currently in esri-leaflet-doc: https://github.com/Esri/esri-leaflet-doc/blob/master/data/integrity.js. It is slightly different than that one though: instead of relying on the packages being installed via NPM, it makes the requests to find the "latest version" from NPM. 

The format of the `siteData.json` is also slightly updated to be more consistent. (see [old](https://raw.githubusercontent.com/Esri/esri-leaflet-doc/f2379361999bf46cd0579fe3d0880a32746331ee/data/siteData.json) vs [new](https://raw.githubusercontent.com/Esri/esri-leaflet/45b3440950e55a08e1548699f08c4e0949c62a6c/siteData.json)).

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -7,6 +7,9 @@ NAME=$(node --eval "console.log(require('./package.json').name);")
 # build and test
 npm test || exit 1
 
+# update integrity
+npm run integrity
+
 # checkout temp branch for release
 git checkout -b gh-release
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -30,6 +30,7 @@
         "leaflet": "^1.6.0",
         "mkdirp": "^1.0.4",
         "mocha": "^8.4.0",
+        "node-fetch": "^2.6.7",
         "npm-run-all": "^4.1.5",
         "rollup": "^2.49.0",
         "rollup-plugin-uglify": "^6.0.4",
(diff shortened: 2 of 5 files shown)


jwasilgeo (approved):
Looks solid! Checks out fine for me on node v16.13.2 and npm v6.14.11.

scripts/release.sh
```diff
@@ -7,6 +7,9 @@ NAME=$(node --eval "console.log(require('./package.json').name);")
 # build and test
 npm test || exit 1
 
+# update integrity
+npm run integrity
```
patrickarlt, line 11: This runs BEFORE the NPM publish so it won't get the integrity data for the latest version. I think this has to run AFTER `npm publish` on line 37 as the last thing in the scripts.

patrickarlt (changes requested):
Looks good but I think we need one change.

scripts/release.sh
gavinr: Thanks Patrick, this is a good discussion: If we move it to AFTER the `npm publish`, then we'll also have to make sure we do a final additional commit on the master branch and then it would be available at `https://github.com/Esri/esri-leaflet/blob/master/siteData.json` - I'll update `release.sh` to do that. 

(I'm also a bit concerned about the timing - not sure if the NPM API will immediately start reporting the new version number or if it takes a few seconds because of CDN caching  - I guess we'll just have to try and see though)

PR #1313, opened 2022-02-14
added siteinfo step to release

gavinr:
Adds an additional step to the release bash script that generates the ssri integrity string and saves it (and the lib version number) into a JSON file in the root of the repository (`siteData.json`). This is generating the integrity string based on the built file on disk (as opposed to reading it from NPM) for the security/integrity of the process.

We will do a similar thing in all the related repositories (`esri-leaflet-geocoder`, `esri-leaflet-vector`, etc) so that it will replace the functionality that https://github.com/Esri/esri-leaflet-doc/blob/master/data/integrity.js is currently providing. 

I got the command from [MDN: Subresource Integrity](https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity#tools_for_generating_sri_hashes). It does require OpenSSL to be installed on the build machine but I think that's fairly common/standard.

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -9,7 +9,7 @@ npm test || exit 1
 
 # Integrity string and save to siteData.json
 JS_INTEGRITY=$(cat dist/esri-leaflet.js | openssl dgst -sha512 -binary | openssl base64 -A)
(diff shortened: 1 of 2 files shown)


patrickarlt (changes requested):
LGTM but do you want to write the integrity file to `dist/siteData.json` so it becomes part of the release files?

gavinr:
Thanks @patrickarlt. Sounds good - done.

PR #1317, opened 2022-02-23
update npm release process

gavinr:
In today's builds, `siteData.json` is still [not getting published to npm](https://unpkg.com/browse/esri-leaflet@3.0.6/dist/). 

This is because we have an issue with our release process: right now, `release.sh` generates the dist files [on line 8](https://github.com/Esri/esri-leaflet/blob/6a8b7db704e3086b8b25dbd2d46bcd4d8f6b35a4/scripts/release.sh#L8) when `npm test` runs. Then later when [`npm publish` runs](https://github.com/Esri/esri-leaflet/blob/6a8b7db704e3086b8b25dbd2d46bcd4d8f6b35a4/scripts/release.sh#L38), [`npm prepare` runs as part of that](https://docs.npmjs.com/cli/v8/using-npm/scripts#life-cycle-scripts), which [runs `npm run build`](https://github.com/Esri/esri-leaflet/blob/6a8b7db704e3086b8b25dbd2d46bcd4d8f6b35a4/package.json#L85), _re-building the dist files_. I think this is unnecessary and causing `siteData.json` to not be included in the npm release.

So, this PR changes it so that when publishing, do not use "prepare" script to re-build the dist files, just use the files that are there from earlier in `release.sh` when it ran `npm test`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -82,7 +82,6 @@
     "build": "rollup -c profiles/debug.js & rollup -c profiles/production.js",
     "lint": "semistandard | snazzy",
     "prebuild": "mkdirp dist",
-    "prepare": "npm run build",
     "pretest": "npm run build",
     "precommit": "npm run lint",
     "fix": "semistandard --fix",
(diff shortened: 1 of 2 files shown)


patrickarlt:
LGTM! Does `siteData.json` also get included in the GitHub release?

Issue #1274, opened 2021-05-24
"Accessing non-existent property" warnings


gavinr:
## Describe the bug

Seeing warnings when I run `npm run test`. I do not think it used to happen. I suspect it has to do with my updated version of Node or NPM.

```bash
husky > pre-commit (node v14.15.5)

> esri-leaflet@3.0.2 lint
> semistandard | snazzy

(node:28284) Warning: Accessing non-existent property 'cat' of module exports inside circular dependency
(Use `node --trace-warnings ...` to show where the warning was created)
(node:28284) Warning: Accessing non-existent property 'cd' of module exports inside circular dependency
(node:28284) Warning: Accessing non-existent property 'chmod' of module exports inside circular dependency
(node:28284) Warning: Accessing non-existent property 'cp' of module exports inside circular dependency
(node:28284) Warning: Accessing non-existent property 'dirs' of module exports inside circular dependency
(node:28284) Warning: Accessing non-existent property 'pushd' of module exports inside circular dependency
(node:28284) Warning: Accessing non-existent property 'popd' of module exports inside circular dependency
(node:28284) Warning: Accessing non-existent property 'echo' of module exports inside circular dependency
(node:28284) Warning: Accessing non-existent property 'tempdir' of module exports inside circular dependency
(node:28284) Warning: Accessing non-existent property 'pwd' of module exports inside circular dependency
(node:28284) Warning: Accessing non-existent property 'exec' of module exports inside circular dependency
(node:28284) Warning: Accessing non-existent property 'ls' of module exports inside circular dependency
(node:28284) Warning: Accessing non-existent property 'find' of module exports inside circular dependency
(node:28284) Warning: Accessing non-existent property 'grep' of module exports inside circular dependency
(node:28284) Warning: Accessing non-existent property 'head' of module exports inside circular dependency
(node:28284) Warning: Accessing non-existent property 'ln' of module exports inside circular dependency
(node:28284) Warning: Accessing non-existent property 'mkdir' of module exports inside circular dependency
(node:28284) Warning: Accessing non-existent property 'rm' of module exports inside circular dependency
(node:28284) Warning: Accessing non-existent property 'mv' of module exports inside circular dependency
(node:28284) Warning: Accessing non-existent property 'sed' of module exports inside circular dependency
(node:28284) Warning: Accessing non-existent property 'set' of module exports inside circular dependency
(node:28284) Warning: Accessing non-existent property 'sort' of module exports inside circular dependency
[...]

PR #1320, opened 2022-02-25
updated to semistandard 12

gavinr:
The latest version of semistandard is v16. We have not been keeping up to date but we now need to get up to date due to dependency security issues. So we'll tackle these one major version at a time. Here are the code changes required to get up to v12.

I think they're all just whitespace fixes.


spec/Layers/DynamicMapLayerSpec.js
```diff
@@ -523,10 +523,10 @@ describe('L.esri.DynamicMapLayer', function () {
     });
 
     var customIdentify = L.esri.identifyFeatures({ url: url })
-                          .layers('all:0')
```
jwasilgeo, line 526: 👨‍🍳 👍

jwasilgeo (approved):

spec/Layers/DynamicMapLayerSpec.js
jwasilgeo: 🚫 👾

gavinr:
Bonus: this fixes #1274!

PR #1325, opened 2022-03-23
fix warning in Leaflet 1.8.0

patrickarlt:
When an overlay loads we should remove the error handler not the already removed load handler.


gavinr:
tested: fixes the "listener not found" when using with https://unpkg.com/leaflet@1.8.0-beta.0/dist/leaflet.js with  `L.esri.dynamicMapLayer`.

gavinr:
This was released in v3.0.8

Issue #1335, opened 2022-06-27
Switch from karma-chai-sinon to karma-sinon-chai


gavinr:
### Describe the problem

This is the package we're currently using:

- [karma-chai-sinon](https://www.npmjs.com/package/karma-chai-sinon) - 0.1.5 • Published 7 years ago

... seems like it is not well maintained.

### Describe the proposed solution

It seems like this fork is better maintained:

- [karma-sinon-chai](https://www.npmjs.com/package/karma-sinon-chai) - 2.0.2 • Published 4 years ago

### Additional Information

When I switch to karma-sinon-chai [here](https://github.com/Esri/esri-leaflet/blob/e1ae96d745bbc070d65c2949a48d9f897b1c8951/karma.conf.js#L12) though, the number of tests that get run goes from 311 to 253, so there's an issue there that we need to investigate why switching causes a drop in test cases that are run.

Issue #1318, opened 2022-02-24
Automated testing using GitHub Actions


gavinr:
### Describe the problem

Travis.org automated testing stopped working in June 2021: https://travis-ci.org/github/Esri/esri-leaflet ... so we have no automated testing on commits or pull requests.

### Describe the proposed solution

We should update to some other solution, like GitHub Actions or Travis-ci.com.

### Additional Information

Once a good solution is found, we should do the same setup on all the other esri-leaflet-* repos.

PR #1336, opened 2022-06-27
Automated testing with GitHub actions

gavinr:
- Updated karma config to keep it in line with Leaflet (https://github.com/Leaflet/Leaflet/blob/a3f1ba930536c036bfd5890dbae5fb18e8ee82d8/spec/karma.conf.js)
- Removed TravisCI configuration and enabled GitHub Actions to automatically test our code when it is pushed to GitHub (via `.github/workflows/main.yml`). (#1318). Follows the example of Leaflet: https://github.com/Leaflet/Leaflet/blob/a3f1ba930536c036bfd5890dbae5fb18e8ee82d8/.github/workflows/main.yml . 
    - This automatically tests our code in Chrome, Firefox, and Edge.
- Switched to use `karma-sinon-chai` since it seems to be better maintained (#1335)


patrickarlt (approved):
Can't argue with this. 👍🏻

Issue #1342, opened 2022-09-22
Support for Leaflet 1.9.0


addy:
### Describe the problem

Looks like `esri-leaflet` is currently on the 1.6.0 release of Leaflet. There have been a few changes, improvements and bug fixes since then. Does Esri have a usual timeline for incorporating new versions of Leaflet?

### Describe the proposed solution

Libraries built on top of Leaflet that can be used in conjunction with Esri's plugin use more advanced versions of Leaflet. It would be great if this library supported the latest versions under the hood and allowed consumers to be able to upgrade their Leaflet dependencies.

jgravois:
since leaflet is a peerDependency you can install any version >= 1.0.0 that you want.

https://github.com/Esri/esri-leaflet/blob/7ee4b61140e160f3b27bcc604be5b79829f2fb9a/package.json#L77

if you're seeing a specific problem here because of a breaking change in leaflet, it'd helpful if you could describe it in detail.

addy:
Sure, yeah here's the specific issue that I'm seeing:

```
Failed to compile.

./node_modules/esri-leaflet/src/Layers/FeatureLayer/FeatureGrid.js
Attempted import error: 'toLatLngBounds' is not exported from 'leaflet' (imported as 'latLngBounds').
```

gavinr:
@addy thanks, can you please post the steps I would need to follow to see that error?

jgravois:
i assume you'll see what @addy mentioned as soon as you `npm run build`.

gavinr:
Thanks @jgravois for the help. That workflow actually works for me:
![image](https://user-images.githubusercontent.com/209355/191887295-9a21daf6-5697-4fb8-ade4-0dee33deae66.png)

addy:
I'll work on getting a StackBlitz, or similar, up shortly to reproduce the error I'm seeing locally. I think it's likely an issue arising from a mix of dependencies in an external project using both Leaflet and esri-leaflet.

addy:
@jgravois @gavinr Alright I've recreated the issue here: https://stackblitz.com/edit/vitejs-vite-k5qazv

Let me know if you have issues accessing that.

gavinr:
Thanks for the reproduction case. I do see it there, and have also tested on this project: https://github.com/gavinr/esri-leaflet-react-demo/tree/leaflet-1.9 and get the same error:
[...]

PR #1343, opened 2022-09-28
Fix toLatLngBounds was not found in 'leaflet'

fuhidev:
export 'toLatLngBounds' (imported as 'latLngBounds') was not found in 'leaflet'

### DIFF
--- a/src/Layers/FeatureLayer/FeatureGrid.js
+++ b/src/Layers/FeatureLayer/FeatureGrid.js
@@ -1,6 +1,6 @@
 import {
   LatLngBounds,
-  toLatLngBounds as latLngBounds,
+  latLngBounds,
   Layer,
   Browser,
   Util,


gavinr:
Thank you for the PR!. This is a potential fix for #1342. We'll test and merge this soon.

pavch:
waiting eagerly this fix

gavinr:
Hi, we are hoping that the fixes happening in leaflet (https://github.com/Leaflet/Leaflet/issues/8451 / https://github.com/Leaflet/Leaflet/pull/8493) will resolve this without this PR being required. If the leaflet 1.9 patch does not resolve it, we will look more into this solution in the PR. Thanks for your patience!

pavch:
but what about compatibility with older versions of Leaflet?

addy:
@gavinr I saw Leaflet released `1.9.1` which doesn't seem to fix the issue :\

gavinr:
Leaflet released v1.9.2 a few hours ago: https://github.com/Leaflet/Leaflet/releases/tag/v1.9.2

And [I have tested](https://github.com/gavinr/esri-leaflet-react-demo/pull/2) that using 1.9.2 (instead of 1.9.0 or 1.9.1) fixes the issue #1342, so I think the solution is for anyone who wants to use leaflet v1.9 to use >=1.9.2 .... and this PR can be closed without merging. I will close this soon unless I hear any major objections.

Thanks!

gavinr:
I am assuming this is resolved if using leaflet v1.9.**2** (https://github.com/Esri/esri-leaflet/issues/1342#issuecomment-1267555419). If not, please comment or open a new issue. Thank you!

mbostock:
This still seems to be broken. Here is a simple reproduction using jsDelivr’s +esm service:

```js
import("https://cdn.jsdelivr.net/npm/esri-leaflet/+esm")
```

This generates an error:

```js
SyntaxError: The requested module '/npm/leaflet@1.9.4/+esm' does not provide an export named 'toLatLngBounds'
```

gavinr:
@mbostock thanks for digging up this issue to report this item. I did a bit of research here:

I **do not** get this error when installing via NPM and using the ES modules that way, for example here: https://github.com/gavinr-maps/esri-leaflet-react-demo. Is that the same for you?

So then the issue is when using ES modules via `<script type="module"` with these common CDNs (e.g. Skypack, esm.sh, or jsdeliver.net.../+esm). I can confirm that I **am** getting the same error as you when using those. Replication cases:
- **esm.sh**: https://jsbin.com/nuwoqiyiku/4/edit?html,output - error `Uncaught SyntaxError: The requested module '/v135/leaflet@1.9.4/es2022/leaflet.mjs' does not provide an export named 'toLatLngBounds'`
- **cdn.jsdelivr.net...../+esm** - https://jsbin.com/nuwoqiyiku/6/edit?html,output - error `Uncaught SyntaxError: The requested module '/npm/leaflet@1.9.4/+esm' does not provide an export named 'toLatLngBounds'`
- **Skypack**: https://jsbin.com/nuwoqiyiku/2/edit?html,output - works fine, oddly. But [skypack](https://github.com/skypackjs/skypack-cdn/issues/329) [is](https://github.com/skypackjs/skypack-cdn/issues/304) [abandoned](https://github.com/skypackjs/skypack-cdn/issues/365) so I think the above 2 errors are more indicative of the problem.

So I do think this is a _slightly_ different (but similar) issue than the original issue here... It would probably be good to open a **new issue**, focusing on the error when using  ESM url CDNs.

PR #1304, opened 2021-12-18
Fix FeatureLayer.refresh for moved point layers

patrickarlt:
This fixes `FeatureLayer.refresh()` for layers where `simplifyFactor` is not set. Originally this was intended to update the geometry when a higher resolution geometry came in from the service but it also means that layers with edited geometries were missed by `FeatureLayer.refresh()`

Can be reproudced with the following HTML in the `debug` folder (originally from @gavinr):

```html
<!-- https://esri.github.io/esri-leaflet/examples/simple-editing.html -->
<!-- Forked from https://codepen.io/jwasilgeo/pen/KKXVjvy?editors=1000 -->
<!DOCTYPE html>
<html>

<head>
  <meta charset="utf-8" />
  <title>Editing feature layers</title>
  <meta name="viewport" content="initial-scale=1,maximum-scale=1,user-scalable=no" />

  <!-- Load Leaflet from CDN -->
  <link rel="stylesheet" href="https://unpkg.com/leaflet@1.7.1/dist/leaflet.css" />
  <script src="https://unpkg.com/leaflet@1.7.1/dist/leaflet.js"></script>

  <!-- Load Esri Leaflet from source-->
  <script src="../dist/esri-leaflet-debug.js"></script>

  <style>
    body {
      margin: 0;
      padding: 0;
    }

    #map {
      position: absolute;
      top: 150px;
      bottom: 0;
      right: 0;
      left: 0;
    }
  </style>
</head>

<body>

  <!-- Leaflet Draw -->
  <style>
    #info-pane {
      z-index: 400;
      padding: 1em;
      background: white;
      text-align: right;
    }

    #form {
      display: none;
    }
  </style>

  <div id="map"></div>
  <div id="info-pane" class="leaflet-bar" style="font-size: 10px">
    These do not use the Leaflet or Esri Leaflet add/edit funcitonality. They call the REST endpoints using FETCH
    directly:<br />
    <button id="addFeatureButton">ADD a feature within the current extent</button><br />
    <button id="moveFeatureButton">MOVE one of the features within the current extent</button>
    <br /><br />
    This DOES call the leaflet layer.refresh() function:<br />
    <button id="refreshLayerButton">Refresh the layer</button>
  </div>

  <script>
    // create the map
    var map = L.map('map').setView([38.631, -90.199], 8);
    // basemap
    var OpenStreetMap_Mapnik = L.tileLayer('https://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png', {
      maxZoom: 19,
[...]

PR #1346, opened 2022-12-05
Fix performance issue with refreshing features

### LINKED ISSUE #1340: featureLayer rendering performance
### Describe the bug

Hi, i've noticed that any attempt to style a feature layer, results in a massive overall map rendering performance hit. Earlier versions of esri-leaflet.js, e.g. 2.5.3 work as expected.

Thanks.

### Reproduction

Here's a simple code snippet:

 var vGrassStyle = {
        "color": "#00ff00"
        };

 var vLayer = L.esri.featureLayer({
        url: 'https://services6.arcgis.com/xjnLSXTALbkc5QYN/arcgis/rest/services/Grass_Cutting/FeatureServer/0',
        style: vGrassStyle
        });

### Logs

_No response_

### System Info

```shell
Leaflet 1.8.0, ESRI 3.0.8
```

### Additional Information

_No response_

patrickarlt:
This should resolve #1340 which is a result of an error I made in #1304.

1. In #1304 I added a call to `redraw`.
2. Internally this eventually called [`resetStyle`](https://github.com/Esri/esri-leaflet/blob/master/src/Layers/FeatureLayer/FeatureLayer.js#L410)
3. `resetStyle` calls `resetFeatureStyle` on every feature in the layer. This ends up causing a huge amount of extra calls since it calls since the happens `N * N` times.

This PR simply changes this to call `resetFeatureStyle` directly for the specific feature.

### DIFF
--- a/src/Layers/FeatureLayer/FeatureLayer.js
+++ b/src/Layers/FeatureLayer/FeatureLayer.js
@@ -407,7 +407,7 @@ export var FeatureLayer = FeatureManager.extend({
 
     // looks like a path (polygon/polyline)
     if (layer && layer.setStyle && this.options.style) {
-      this.resetStyle(geojson.id);
+      this.resetFeatureStyle(geojson.id);
     }
   }
 });


gavinr (approved):
Looks great! I tested the performance issue from #1340 and it does seem to fix that case.

gavinr:
This has been released in v3.0.9
https://jsbin.com/qubokaculi/1/edit?html,output is the demo case using v3.0.9 and it is working.

PR #1351, opened 2023-01-11
Update dependencies

gavinr:
- Updated misc dependencies
- removed `uglify-js` - we're not using that anymore, it should have been removed in https://github.com/Esri/esri-leaflet/pull/1316
- switching from `rollup-plugin-terser` to `@rollup/plugin-terser` (see https://github.com/TrySound/rollup-plugin-terser/issues/123)


patrickarlt (approved):
LGTM!

PR #1360, opened 2023-05-07
src/Util.js - custom Leaflet prefix support

markconnellypro:
When Esri Leaflet adds _Powered by [Esri](http://www.esri.com/)_ to the attribution, it overwrites the preexisting attribution prefix. This pull request changes this process to retrieve the existing attribution prefix first, so that it can be used in lieu of the constant when overwriting and so that it can be restored in the event that all Esri layers are moved from the map display.

### DIFF
--- a/spec/Layers/BasemapLayerSpec.js
+++ b/spec/Layers/BasemapLayerSpec.js
@@ -110,6 +110,29 @@ describe('L.esri.BasemapLayer', function () {
     expect(L.esri.basemapLayer('Topographic')).to.be.instanceof(L.esri.BasemapLayer);
   });
 
+  it('should have correct attribution when attribution is set BEFORE adding the layer to the map', function () {
+    map.attributionControl.setPrefix("ccc");
+    const layer = L.esri.basemapLayer("Topographic");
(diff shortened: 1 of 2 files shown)


gavinr:
Thanks for the PR. Your explanation sounds reasonable, but also I want to make sure we have a full discussion and thoughtful understanding of the problem and solution first. Given that, would you mind [opening an issue](https://github.com/Esri/esri-leaflet/issues/new/choose) for this (and you can refernce this PR within that issue),  clarifying the issue and the replication cases? Thank you again!

markconnellypro:
I pushed a commit to address the empty string issue (item 2) -- hopefully this is sufficient to fix those tests.

With respect to overwriting Powered by Esri (item 1), this appears to be a preexisting issue independent of this pull request. This is a bit at the border of my JS knowledge, but I assume to get this to work that Esri Leaflet would have to override the setPrefix method in the Control.Attribution class in Leaflet? I'm also not sure what impact such an override would have on the other functions in src/Util.js that call the setPrefix method.

markconnellypro:
Closing this pull request due to #1366.

PR #1372, opened 2023-09-25
Only load featurelayer tiles after metadata has been loaded

markhepburn:
This can have a significant impact on performance for larger layers, where the initial detailed tiles are requested concurrently with the metadata request and thus loaded in arcgis format -- assuming isModern hasn't been manually set -- and need to be converted to geojson (which can choke a browser).

This is only a partial fix, as _setView can also trigger  _update.

A potential fix for this is an instance variable to check whether metadata has loaded yet (tricky as metadata is requested in a subclass), or some proxy for this -- I have been using "_cells is empty or not", but there may be a better way, and I'm possibly missing cases as well.

### DIFF
--- a/src/Layers/FeatureLayer/FeatureGrid.js
+++ b/src/Layers/FeatureLayer/FeatureGrid.js
@@ -48,7 +48,6 @@ export var FeatureGrid = Layer.extend({
     this._cells = {};
     this._activeCells = {};
     this._resetView();
-    this._update();
   },
 
   onRemove: function (map) {
--- a/src/Layers/FeatureLayer/FeatureManager.js
+++ b/src/Layers/FeatureLayer/FeatureManager.js
@@ -111,6 +111,8 @@ export var FeatureManager = FeatureGrid.extend({
           map.attributionControl.addAttribution(this.getAttribution());
         }
       }
+      // Only request cells once we know whether we have geoJSON support or not:
+      this._update();
     }, this);
 
     map.on('zoomend', this._handleZoomChange, this);


gavinr-maps:
Thanks for the PR. Do you have some sample code that would demonstrate the issue you're fixing here?

patrickarlt (commented):
This fix looks good on the surface but it means that sub-classes of `FeatureGrid` are now responsible for calling the initial `_update` when they are ready to start loading features.

The issue is that there are TONS of cases were things will happen that will start loading features before we are ready. For example what if the user pans or zooms after loading a layer but before we know if it supports GeoJSON or not?

I'll have to give a think on this. I'm leaning towards just making `isModern` enabled by default now. `f=geojson` has been supported since ArcGIS Server 10.4 ([source](https://community.esri.com/t5/developers-blog/new-geojson-capabilities-for-developers/ba-p/892090)) which was officially deprecated in February 2022 ([lifecycle doc](https://support.esri.com/en-us/products/arcgis-enterprise/life-cycle)) so I think we could probably make GeoJSON the default now and avoid this issue.

This would however be a breaking change and require a major version bump.

markhepburn:
@patrickarlt Thanks for looking at this, that all makes sense.

One thing that would still be affected is the Esri/esri-leaflet-renderers; I found that early polygons are unstyled, until you zoom in or otherwise load more.  It's a tricky problem though, that's why I only called this a partial fix!

patrickarlt:
@markhepburn closing this in favor of https://github.com/Esri/esri-leaflet/issues/1376. Thanks for bringing this up. In the meantime I would just manually set `isModern`.

Yes @gavinr-maps and I decided to archive `Esri/esri-leaflet-renderers` since it introduces a lot of additional support requests and expands the scope of the library with the expectation that we will support lots of other things which we really don't have the capacity for.

markhepburn:
@patrickarlt Quite understand, thanks for taking the time to look at this.

What's the current recommended platform for incorporating esri feature service layers (with styling), out of curiosity?  We picked Leaflet for various reasons including excellent React bindings, but it's getting harder to support non-image formats such as ESRI or pbf.

patrickarlt:
> What's the current recommended platform for incorporating esri feature service layers (with styling), out of curiosity?

This would be the official [ArcGIS Maps SDK for JavaScript](https://developers.arcgis.com/javascript/latest/) which supports everything you can configure in the map viewer and ArcGIS Pro and as a bonus also gives you full web map support which is too large in scope for us to build for Esri Leaflet. Keep and eye out for some announcements this month with regard to some new releases which should hopefully make integration with 3rd party frameworks easier.

Issue #1369, opened 2023-08-07
Husky update


gavinr-maps:
### Describe the bug

We are getting a warning during the build:
![image](https://github.com/Esri/esri-leaflet/assets/137905994/6e459376-50b4-4f64-b50b-40c65ccd8593)

### Reproduction

` git commit -m "build $VERSION"`

### System Info

```shell
Latest
```

Richarizaard:
@gavinr-maps I created #1374 to update how we define our pre-commit in our package.json.

PR #1374, opened 2023-09-27
Update husky pre-commit hook in package.json

Richarizaard:
#1369 

Updates how we set our pre-commit in package.json.

Moved from `scripts` to `husky.hooks`.

image


gavinr-maps (approved):
thanks!

PR #1377, opened 2023-10-10
Url encode apostrophes in request params

### LINKED ISSUE #1246: Request URL length <= 2000 checking does not escape `'`
Tested on Leaflet 1.7.1, Esri Leaflet 2.5.1.

I've encountered a issue where a request to a Feature layer through resource proxy would fail with 404 error when filter condition reaches specific length. Looking at #983, I found that `Request.js` is calculating the length of the request to be less than the actual length, and it just so happens hit the 2000 (probably 2048?) character limit so it's incorrectly using GET request, which caused the fail.

https://github.com/Esri/esri-leaflet/blob/master/src/Request.js#L121

With a debugger I see that `paramString` did not escape `'` to `%27` while the actual request URL would, hence under-estimating the URL length.

BrunoCaimar:
Fix #1246

Replace the apostrophe sign ' with %27

### DIFF
--- a/spec/Layers/FeatureLayer/FeatureManagerSpec.js
+++ b/spec/Layers/FeatureLayer/FeatureManagerSpec.js
@@ -652,7 +652,7 @@ describe('L.esri.FeatureManager', function () {
   it('should filter features with a where parameter', function () {
     server.respondWith(
       'GET',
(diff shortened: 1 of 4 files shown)


gavinr-maps (approved):
Looks good. Thanks!

PR #1378, opened 2023-10-11
Add a hideToken option to avoid sending the token via querystring

### LINKED ISSUE #1364: Token is exposed in GET request
### Describe the bug

When connecting to a map service that requires a token, creating a layer exposes the token, which is a sensitive parameter, as part of a GET request. Is there any way to make it a POST request instead?

### Reproduction

To reproduce this issue, you will need a service that requires the use of a token. Open your preferred web browser and inspect the network traffic. You should see a GET request being made to the service which includes the token in the url.

Here's the code I use to create a dynamic map layer. I cannot disclose the value of baseUrl here.

```
L.esri.dynamicMapLayer({
            url: `${baseUrl}/server/rest/services/CC_SNET_Boundary/CC_SNET_HUB_BNDY_DYN/MapServer`,
            opacity: 1,
            useCors: true,
            token: token
        });
```

### Logs

_No response_

### System Info

```shell
Leaflet version: 1.0.1 
Esri Leaflet version: 2.2.3
```

### Additional Information

_No response_

BrunoCaimar:
+ Our security scanning does not allow sending tokens via GET anymore
+ Sending the token via the request header does not work because map services do not allow the preflight request 

This can potentially fix #1364.

### DIFF
--- a/spec/RequestSpec.js
+++ b/spec/RequestSpec.js
@@ -48,6 +48,38 @@ describe('L.esri request helpers', function () {
     requests[0].respond(200, { 'Content-Type': 'text/plain; charset=utf-8' }, JSON.stringify(sampleResponse));
   });
 
+  it('should not switch from a GET request if hideToken option is on, but no token is present', function (done) {
+    L.esri.get.CORS('http://services.arcgisonline.com/ArcGIS/rest/info', {}, function (error, response) {
+      expect(this.foo).to.equal('bar');
+      expect(response).to.deep.equal(sampleResponse);
+      done();
+    }, {
+      foo: 'bar',
+      options: { hideToken: true }
+    });
+
+    expect(requests[0].url).to.equal('http://services.arcgisonline.com/ArcGIS/rest/info?f=json');
+    expect(requests[0].method).to.equal('GET');
+    requests[0].respond(200, { 'Content-Type': 'text/plain; charset=utf-8' }, JSON.stringify(sampleResponse));
+  });
+
+  it('should be able to switch from a GET request to POST if token is present and hideToken option is on', function (done) {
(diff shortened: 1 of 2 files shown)


patrickarlt:
@BrunoCaimar I took the day talk this through with internal teams at Esri. The consensus is that this behavior (token in the query string for `GET` requests) isn't really something we want to work around by making everything `POST` requests. The main issue is that `POST` requests are not cached anywhere in the system and this would defeat the work we have some to make feature layer requests tileable, cacheable and repeatable for the best performance. In the case of viral or heavily accessed feature services you would almost certainly run into rate limiting.

In order to hide the token in `GET` requests and still have the queries be cacheable we would need the server to properly handle preflighting `OPTIONS` requests. This could be done but it means that you double the number of requests and introduce another 150-200ms request-response cycle to do the preflightling which was judged as an unacceptable performance hit.

To mitigate this you can do the following:

* If you are accessing private user data use short lived tokens via oAuth 2.0 with PKCE. This will mean that tokens that do exist in logs are short lived in will likely be invalid by the time they could be used maliciously. This is the approach taken by the ArcGIS Maps SDK for JavaScript by default.
* If you are using API keys we recommend rotating keys on a regular basic for the same reasons above. We recommend this in [our best practices](https://developers.arcgis.com/documentation/mapping-apis-and-services/security/security-best-practices/#api-key-security)
* If you using API Keys and want shorter expiration times for tokens you could switch to [app credential authentication](https://developers.arcgis.com/documentation/mapping-apis-and-services/security/app-credential-authentication/) and generate a short lived token on the server to use on the client. Since these tokens expire quickly you can simply generate as many tokens as you need with short expiration.

PR #1390, opened 2024-05-15
Fix use of timeout property for synchronous XMLHttpRequests

bscuron:
Convert XMLHttpRequests to be asynchronous to allow for the use of the timeout property. The timeout property can't be used for synchronous requests in a document environment (window/iframe) or an `InvalidAccessError` is thrown. (Please see the [documentation](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/timeout))

This issue only appeared after updating my Salesforce org to the Summer '24 Release. I am deploying the Esri Leaflet map in Salesforce with LWC.

All `npm test` tests pass with the new changes.

**Browser:** Chrome
**Browser version:** 124.0.6367.207
**Esri Leaflet version:** 3.0.12
**Leaflet version:** 1.9.4
**Bundling tool:** Salesforce

### DIFF
--- a/src/Request.js
+++ b/src/Request.js
@@ -123,9 +123,9 @@ export function request (url, params, callback, context) {
 
   // ie10/11 require the request be opened before a timeout is applied
   if (requestLength <= 2000 && Support.cors) {
-    httpRequest.open('GET', url + '?' + paramString);
+    httpRequest.open('GET', url + '?' + paramString, true);
   } else if (requestLength > 2000 && Support.cors) {
-    httpRequest.open('POST', url);
+    httpRequest.open('POST', url, true);
     httpRequest.setRequestHeader('Content-Type', 'application/x-www-form-urlencoded; charset=UTF-8');
   }
 


gavinr-maps:
Hi @bscuron, thank you for the PR. I'm a bit confused why this is needed here - the MDN documentation for `XMLHttpRequest.open` says that the async property is `true` by default:
image
https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/open

... so it seems like this PR is not changing anything. Could you maybe expand a bit more and provide a minimal demonstration case of the problem you're running into? Thanks!

bscuron:
Hi @gavinr-maps, thank you for the response.

The Salesforce Summer '24 release most likely has a defect during LWC compilation where the third argument is not being treated as optional. By explicitly setting the third argument as `true` in `open()`, the library behaves as it did prior to the Salesforce Summer '24 release that occurred on May 11.

I plan to submit a support ticket to Salesforce to get the issue resolved. Would you consider merging this PR, since it does not affect users of the library in any negative way, but fixes an issue for users using the Salesforce Summer '24 release?

jgravois:
just my (unsolicited) two cents, but i'd consider a situation like this an ideal scenario to bundle a customized version of the library for yourself.

obviously there's overhead involved in tracking upstream changes any time you do this, but this library is pretty darn stable at this point.

patrickarlt:
@bscuron I think I agree with @jgravois here. Esri Leaflet is fairly stable at this point and it seems like this works in all of our target environments. If this is indeed a bug in Salesforce they should fix it and in the meantime you could use a custom build of this library until then.

PR #1394, opened 2024-10-24
Remove usage of toLatLngBounds

### LINKED ISSUE #1386: Error when using ESM CDNs - `The requested module does not provide an export named 'toLatLngBounds'`
### Describe the bug

When trying to load Esri Leaflet via ES Modules via an ESM CDN like [esm.sh](https://esm.sh/) or [esm.run by jsDelivr](https://www.jsdelivr.com/esm), there is an error:

> Uncaught SyntaxError: The requested module .... does not provide an export named 'toLatLngBounds' ...

#### Notes:
- This is possibly related or a duplicate of #1343 
- This is similar or the same as what was discussed [here](https://github.com/Esri/esri-leaflet/issues/1342#issuecomment-1258992755).
- This was recently re-reported by @mbostock [here](https://github.com/Esri/esri-leaflet/pull/1343#issuecomment-1906806585).
- This is also tangentially related to #1338

#### esm.run reproduction:

1. Open https://jsbin.com/xokuzam/1/edit?html,output
2. View the developer console
    - Expected: the map loads and no errors
    - Actual:
        > Uncaught SyntaxError: The requested module '/npm/leaflet@1.9.4/+esm' does not provide an export named 'toLatLngBounds' (at +esm:7:265)

### Logs

_No response_

### System Info

```shell
- Leaflet v1.9.4
- Esri Leaflet v3.0.12
```

### Additional Information

It does look like Leaflet exports as `latLngBounds` - https://github.com/Leaflet/Leaflet/blob/b2fd591c33785227e0d4036c1a80a00bd2debd53/src/geo/index.js#L2

... so is the fix as simple as changing this:
https://github.com/Esri/esri-leaflet/blob/48cfd690614594428b09da284e30e08892a8ab6e/src/Layers/FeatureLayer/FeatureGrid.js#L3

to:

```js
latLngBounds,
```

?

gavinr-maps:
Resolves #1386 (also possibly related to #1200 ?)

After this change, ESM CDNs can be used. I published this change to https://www.npmjs.com/package/el-esm-test to be able to test it using https://esm.sh/el-esm-test@3.0.14 ... see here for an example of using that:
https://jsbin.com/xoxopus/1/edit?html,output

Note that using **esm.run** (JSDelivr's ESM CDN): https://esm.run/el-esm-test@3.0.14
... does **not** work. We think this is a bug with esm.run and will possibly follow up with a bug report to that project. For now our recommendation will be to **use esm.sh if you want to use an ESM CDN with esri-leaflet**.

### DIFF
--- a/src/Layers/FeatureLayer/FeatureGrid.js
+++ b/src/Layers/FeatureLayer/FeatureGrid.js
@@ -1,6 +1,6 @@
 import {
   LatLngBounds,
-  toLatLngBounds as latLngBounds,
+  latLngBounds,
   Layer,
   Browser,
   Util,


patrickarlt (approved):
LGTM!

Issue #1384, opened 2024-01-06
MapServer with custom LODs. initial tile set zoom level is wrong


roocell:
### Describe the bug

I have a MapServer with custom zoom levels (4 to 9) which corresponds to 13 to 18 on my map.
I set initial map zoom level to 13 and when I add my tile set it is requesting zoom 13 - which makes sense.
But afterwards (somehow) esri-leaflet figures out that zoom 13 corresponds to 4 on my server - I assume this is in some JSON on the MapServer describing this.
So when i zoom/pan after the initial load it grabs the tiles properly.

how do I get the initial tile load to use zoom 4 rather than 13?

right now, I'm just zoom/panning afterwards to get them to load.

### Reproduction

```
var map = L.map('map').setView([45.39793819727917, -75.72070285499208], 100.0);
map.setZoom(13); // start way out (to prevent so many 404s at startup)

L.tileLayer('https://tile.openstreetmap.org/{z}/{x}/{y}.png', {
    attribution: '&copy; <a href="https://www.openstreetmap.org/copyright">OpenStreetMap</a> contributors'
}).addTo(map);

L.esri.tiledMapLayer({
    url: "https://maps.ottawa.ca/arcgis/rest/services/Basemap_Imagery_1965/MapServer",
    pane: "overlayPane",
    opacity: 1.0, // make all hidden
    maxNativeZoom: 18, // zoom capability of tiles
    maxZoom: 22, // zoom on map (will stretch tiles)
}).addTo(map);
```

### Logs

```shell
initial request
https://maps.ottawa.ca/arcgis/rest/services/Basemap_Imagery_1965/MapServer/tile/13/2935/2370 404 (Not Found)

successful requests after load when zoom/pan
https://maps.ottawa.ca/arcgis/rest/services/Basemap_Imagery_1965/MapServer/tile/4/2937/2371
```

### System Info

```shell
https://unpkg.com/esri-leaflet@3.0.12/dist/esri-leaflet.js
```

gavinr-maps:
`this._lodMap` is the mapping from LODs to zoom levels. If you look at `this._lodMap` [here](https://github.com/Esri/esri-leaflet/blob/e827880907b67aad39c06d1aecc7353b7bd85613/src/Layers/TiledMapLayer.js#L142), it looks correct-- mapping LOD 4 to zoom 13 for example: 

![Image](https://github.com/user-attachments/assets/05cf7ea4-02ca-422e-804b-2c69441d0a4f)

[...]

PR #1400, opened 2025-02-04
potentially fix LODmap issue

gavinr-maps:
#1384


patrickarlt:
@gavinr-maps Do you think it would be possible to come up with a unit test that would reflect this fix?

gavinr-maps:
@patrickarlt since this is in `createTile` which is called from elsewhere in Leaflet, I'm not sure how to test this. I do not see any other unit tests for this `createTile` function either. Could you give me some guidance?

gavinr-maps:
@patrickarlt here is the replication case:

1. Current Behavior: 
    Open https://codepen.io/gavinr/pen/gbpjJMj?editors=1010
    - **Expected**: the imagery layer (black and white layer) should show on load
    - **Actual**: the imagery layer (black and white layer) does not show on load. It does show if you zoom out then zoom back in.

2. Changes from this PR:
    Open https://codepen.io/gavinr/pen/PwqBvBL?editors=1010 
    - **Expected**: the imagery layer (black and white layer) should show on load
    - **Actual**: the imagery layer (black and white layer) should shows on load

patrickarlt (approved):
LGTM

Issue #1376, opened 2023-10-09
Request GeoJSON by deault for feature services


patrickarlt:
### Describe the problem

As discussed in https://github.com/Esri/esri-leaflet/pull/1372#pullrequestreview-1662593212 and https://github.com/Esri/esri-leaflet/issues/1373 there are lots of cases where features are requested before we know if the service supports GeoJSON or not. This introduces to performance issues because browsers have to convert to GeoJSON when we could probably have the service do it.

### Describe the proposed solution

Since GeoJSON output is available since server 10.4 (which is deprecated in Feb 2022) we can probably make `isModern` the default for feature services.

### Alternatives considered

This was discussed in https://github.com/Esri/esri-leaflet/pull/1372 but was deemed to complex to implement.

hiSandog:
Defaulting modern services to GeoJSON could remove avoidable client conversion, but capability uncertainty and proxy behavior still need a fallback path. Could the client optimistically request GeoJSON, cache the service capability, and retry the legacy format only on the specific unsupported-format response? Tests against pre-10.4, modern, and misreporting services would make the performance improvement safe for existing deployments.

Issue #1373, opened 2023-09-25
FeatureLayer: initial tiles are requested concurrently with metadata


markhepburn:
### Describe the bug

FeatureLayers now checks for geojson support, by issuing a metadata query and checking `supportedQueryFormats`.  See #773 and #418 for history.

However, this is done concurrently with the initial set of tiles (`FeatureGrid._update()`), so the first tiles received are still in arcgis format and must be converted to geojson, which can easily choke the browser on large layers.

See #1372 for a partial fix, which calls `_update` at the end of the `metadata` callback, rather than asynchronously.  It's only partial because `_update` is triggered in a few places, so probably needs some way of checking if metadata has loaded yet.

(This all assumes of course that `isModern` has not been manually set, which forces geojson).

### Reproduction

- Monitor browser's network requests for requests to your esri server
- Open https://jsbin.com/weduvomeku/1/edit
- Check the f= parameter in the esri tile requests -- may be slightly deterministic, but expect to see f=json
- Zoom in, to load more tiles
- Check the f= parameter again, which this time should be f=geojson

### System Info

```shell
Leaflet version: v1.9.0
esri-leaflet version v3.0.11

  System:
    OS: Linux 6.2 Ubuntu 22.04.3 LTS 22.04.3 LTS (Jammy Jellyfish)
    CPU: (8) x64 Intel(R) Core(TM) i7-8650U CPU @ 1.90GHz
    Memory: 8.90 GB / 15.36 GB
    Container: Yes
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 20.3.1 - ~/.asdf/installs/nodejs/20.3.1/bin/node
    Yarn: 1.22.19 - ~/.yarn/bin/yarn
    npm: 9.6.7 - ~/.asdf/plugins/nodejs/shims/npm
  Browsers:
    Chrome: 117.0.5938.92
```

alhridoy:
Hi @markhepburn  I've looked into this issue and would like to work on a fix. Is anyone else already addressing this, or can I proceed?

markhepburn:
@alhridoy thanks!

(And I forgot to mention, but this also affects the esri-leaflet-renderers, so initial polygons are often unstyled).

PR #1403, opened 2025-03-22
Fixed: Request GeoJSON by default for feature services 1376

DoiMayank:
The following changes were made:

Updated the logic to default isModern to true for feature services.

Modified instances where isModern is used to ensure GeoJSON is requested by default.

### DIFF
--- a/src/Layers/FeatureLayer/FeatureManager.js
+++ b/src/Layers/FeatureLayer/FeatureManager.js
@@ -80,26 +80,29 @@ export var FeatureManager = FeatureGrid.extend({
         var supportedFormats = metadata.supportedQueryFormats;
 
         // Check if someone has requested that we don't use geoJSON, even if it's available
+       // Check if someone has requested that we don't use geoJSON, even if it's available
         var forceJsonFormat = false;
         if (
-          this.service.options.isModern === false ||
+          this.service.options.isModern === false || // Explicitly disabled
           this.options.fetchAllFeatures
         ) {
           forceJsonFormat = true;
+        } else if (this.service.options.isModern === undefined) {
+          this.service.options.isModern = true; // Default to modern
         }
 
-        // Unless we've been told otherwise, check to see whether service can emit GeoJSON natively
+
+       // Unless we've been told otherwise, check to see whether service can emit GeoJSON natively
         if (
           !forceJsonFormat &&
           supportedFormats &&
           supportedFormats.indexOf('geoJSON') !== -1
         ) {
-          this.service.options.isModern = true;
+          this.service.options.isModern = true; // Confirm GeoJSON support
+        } else if (this.service.options.isModern === undefined) {
+          this.service.options.isModern = true; // Default assumption
         }
 
-        if (metadata.objectIdField) {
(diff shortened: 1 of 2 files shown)


gavinr-maps:
@DoiMayank what issue is this addressing? If this is addressing a bug, can you please provide a replication case for the bug. Thanks!

DoiMayank:
@gavinr-maps This PR addresses the performance issues discussed in #1372 and #1373. The main problem comes from requesting features before confirming if the service supports GeoJSON. This causes unnecessary overhead because browsers have to convert the data to GeoJSON when the service could handle it directly.

The solution is to make isModern the default for feature services since GeoJSON output has been supported in ArcGIS Server starting with version 10.4, which was deprecated in February 2022. By doing this, we avoid the extra processing on the client side and take advantage of the service’s capability to return GeoJSON directly, improving overall performance.

gavinr-maps:
@DoiMayank ok, thanks. So is the replication case in #1373 (copied below) the best test of this fix or something else?
image

DoiMayank:
@gavinr-maps I believe the solution proposed in https://github.com/Esri/esri-leaflet/issues/1376 is the best fix for this issue. I have implemented it in this PR, and @patrickarlt was also leaning in this direction.

gavinr-maps:
Ok, thanks. I'll discuss with Patrick the best way to test this in the context of #1376.

patrickarlt (commented):
@DoiMayank I think this might need some adjustments. In order to trigger the `isModern` behavior down in the service you'll need to set `isModern` to `true` when the service is created in https://github.com/Esri/esri-leaflet/blob/master/src/Layers/FeatureLayer/FeatureManager.js#L40. To do this you'll need to set `isModern: true` in the default options in https://github.com/Esri/esri-leaflet/blob/master/src/Layers/FeatureLayer/FeatureManager.js#L17-L28

After this change `isModern` will default to `true` everywhere and will never be undefined Which should simplify most of your remaining code because we no longer even need to check with the service if it supports GeoJSON or not. We assume it supports GeoJSON unless the user explictly tells us it does not.

If we want to resolve the potential performance issues in https://github.com/Esri/esri-leaflet/issues/1373 then we need to know ahead of time if the service supports GeoJSON which we can reasonably assume it does. Otherwise we have to wait for the metadata call before requesting any features which would be a major rewrite of `FeatureManager`.

patrickarlt:
Another interesting observation from this is that for services that do not support GeoJSON the object id field bust be `OBJECTID` or it likely wont work at all. This is because we don't wait for the metadata call on requests so a non-standard metadata field MIGHT be returned AFTER a query response comes back and we will get some strange behavior when converting to GeoJSON.

patrickarlt:
@DoiMayank I discussed this internally with the team. We all agreed that setting `isModern` to `true` bu default as described in https://github.com/Esri/esri-leaflet/pull/1403#pullrequestreview-2730535236 IF we also print a warning that service does not support GeoJSON and `isModern` is set to `true` the best place for this is probably in https://github.com/Esri/esri-leaflet/blob/master/src/Layers/FeatureLayer/FeatureManager.js#L78-L114

Issue #1404, opened 2025-03-27
Add `apikey` prop to L.esri.TiledMapLayer


MarkTorrey:
### Describe the problem

While working with @sherylgiovanni on some examples for the static basemap tiles service we discovered that L.esri.TiledMapLayer doesn't currently support an `apikey` property. Using `token` does work.

### Describe the proposed solution

Add an `apikey` property to match other implementations such as [L.esri.FeatureLayer](https://developers.arcgis.com/esri-leaflet/api-reference/layers/feature-layer/#options)

### Additional Information

https://codepen.io/MarkTorrey/pen/raNZqLy

gavinr-maps:
This has been released in v3.0.15. Thanks!

PR #1405, opened 2025-03-27
Add tilelayer apikey prop

MarkTorrey:
#1404 - adds an apikey property to L.esri.TiledMapLayer


sherylgiovanni:
cc @gavinr @gavinr-maps @patrickarlt

gavinr-maps (changes requested):
Thanks very much for the PR! It seems to work fine, but on the initial call to the `/mapserver` endpoint it seems like when using `token` it does add it:

image

But with the code in the PR, when using `apikey` parameter it's not included:
image

I think that should probably be consistent, so if you can please update the code to include it in that case too.

MarkTorrey:
closing in favor of a new PR without formatting changes

PR #1407, opened 2025-03-31
Add apikey prop to TiledMapLayer

MarkTorrey:
#1404 - adds an apikey property to L.esri.TiledMapLayer


gavinr-maps (approved):
looks great! Doing is similar to how we're doing it for FeatureLayer:  https://github.com/Esri/esri-leaflet/blob/e827880907b67aad39c06d1aecc7353b7bd85613/src/Layers/FeatureLayer/FeatureLayer.js#L14-L16

Issue #1406, opened 2025-03-31
Use Prettier for code formatting


gavinr-maps:
> @gavinr related to this lets setup prettier as well and remove `semistandard`. 

 _Originally posted by @patrickarlt in [#1238](https://github.com/Esri/esri-leaflet/issues/1238#issuecomment-1278082888)_

Note: After we change this here, we should also change it in the related repos: https://developers.arcgis.com/esri-leaflet/plugins/

gavinr-maps:
Closed with #1408

Issue #1238, opened 2020-11-04
Support Modern (ES6+) JavaScript


gavinr:
In the [source of `esri-leaflet`](https://github.com/Esri/esri-leaflet/tree/master/src), we'd like to start using more modern JavaScript, like [`Object.assign()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign), [arrow functions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions), [template strings](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals), [let](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/let)/[const](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/const), etc.

I'm pretty sure that right now if we try to use any of that, it would cause the library not to work with versions of IE and [per the wiki we should be supporting IE8+](https://github.com//Esri/esri-leaflet/wiki/FAQ#what-browsers-does-esri-leaflet-support). 

So, I'd like to propose changing build process for esri-leaflet to include a transpile step that would allow us to use some of these more modern JavaScript commands within the source, but the build process will allow the code to work with IE8+. It may be as simple as using [@rollup/plugin-babel](https://github.com/rollup/plugins/tree/master/packages/babel) ... not quite sure. Looking for comments and input! Thanks!

jgravois:
> It may be as simple as using @rollup/plugin-babel 

definitely should be. the first commit in https://github.com/Esri/esri-leaflet/pull/1201 contains an example.

gavinr:
Related to all this: the latest `standard`, and thus `semistandard` which we use, has set to give a warning if your project uses `var`.
https://github.com/standard/standard/issues/633#issuecomment-719874016 So we either need to switch to `let` instead of `var`, or be limited to a non-latest version of standard/semistandard.

gavinr:
[...]

PR #1408, opened 2025-04-14
Prettier

gavinr-maps:
#1406 

- Removes semistandard
- Adds ESLint and Prettier
  - Uses the [ESLint Config from the **Leaflet** project](https://github.com/Leaflet/Leaflet/blob/main/eslint.config.js)
- Adds recommended extensions for VS Code

New commands:

- `npm run lint` - tests code for ESLint and Prettier
- `npm run eslint` - tests code for ESLint
- `npm run eslintfix` - fixes ESLint errors
- `npm run prettier` - tests code for Prettier
- `npm run prettierfix` - fixes Prettier errors

There are many code changes in this PR due to the style changes coming along with the ESLint and Prettier changes. I tried to separate the changes out by commit by eslint rule type to make it easier to browse. For example, to see the updates required for the `prefer-rest-params` rule, see https://github.com/Esri/esri-leaflet/commit/0646de1c236d2d573614334e08a5acf4695e465c

Note, I think this also covers #1238 pretty much because the ESLint rules are changing us from var to let/const, arrow functions, etc.


patrickarlt (commented):
@gavinr-maps I think that this looks good overall. I do think that. these rules might be a little complex just because we don't necessarily know all the rules yet.

gavinr-maps:
@patrickarlt ok thanks! Do you think I should make the rules simpler, or are you ok merging as-is? If the latter, I'd like to merge and do a release of Esri Leaflet to ensure there are no regressions coming with these changes.

patrickarlt:
@gavinr-maps lets merge and check for regressions. I think if we feel the rules need changes we can make those decisions later as we work on stuff.

gavinr-maps:
Released in v3.0.16

